PR #30415, opened 2022-02-01
[techdebt]: [PIE-2519]: Add v2 resource for pipeline CRUD with governance checks

namanvermah:
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/30415)

### DIFF
--- a/800-pipeline-service/src/main/java/io/harness/pms/pipeline/PipelineResource.java
+++ b/800-pipeline-service/src/main/java/io/harness/pms/pipeline/PipelineResource.java
@@ -175,7 +175,7 @@ public class PipelineResource implements YamlSchemaResource {
       responses =
       {
         @io.swagger.v3.oas.annotations.responses.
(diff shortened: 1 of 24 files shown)


800-pipeline-service/src/main/java/io/harness/pms/pipeline/PipelineResource.java
```diff
@@ -168,44 +165,6 @@ private PipelineEntity createPipelineInternal(
     return ResponseDTO.newResponse(createdEntity.getVersion().toString(), createdEntity.getIdentifier());
   }
 
-  @POST
```
vaibhav-si, line 171: what if someone is using these APIs? Instead of removing, should we mark them deprecated with some message like "Use v2/pipelines. This will be removed in 14 days" @naidu-harness ?

namanvermah: Added the deprecated annotation

namanvermah:
Will pick this up later

PR #30967, opened 2022-02-24
[fix]: [CI-3582]: Read addon/LE versions only from ci-manager-config.yml

hemanthmantri:
This change is to make the ci-manager-config.yml the authoritative source for addon
and lite-engine versions and not let them overridden by environment or workflow variables.
There will be a separate OPS ticket requesting the removal of environment/workflow variables
for Addon and LiteEngine versions in all the environments.

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`

---
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/30967)


hemanthmantri:
trigger messagecheck

dockerization/ci-manager/scripts/replace_configs.sh
```diff
@@ -48,13 +48,6 @@ if [[ "" != "$NG_MANAGER_URL" ]]; then
   yq write -i $CONFIG_FILE ngManagerClientConfig.baseUrl "$NG_MANAGER_URL"
 fi
 
-if [[ "" != "$ADDON_IMAGE" ]]; then
```
shubham149, line 51: Onprem won't work if this code is removed. Onprem does not have connectivity to dockerhub and downloads the images from private registry.

hemanthmantri: My understanding was that on-prem is handled here:
https://github.com/wings-software/harness-ng-replicated/pull/17

shubham149: replicated sets the ADDON_IMAGE environment variable. Value to set is placed in https://github.com/wings-software/harness-ng-replicated repo.

shubham149: The above code check if this env is set. If set, it replaces the ci-manager config with env value.

hemanthmantri: Okay. So is there any other variable to check if this is ONPREM or SaaS? Can we conditionally allow only ONPREM env to let them override but not SaaS? I don't know if that's even the preferred approach. The goal is to have just one place to change addon/LE version after any modification. Harsh mentioned that with the patch release feature we don't see a need for out-of-band version changes for LE/AddOn.

shubham149: If we don't set the ADDON_IMAGE & LE_IMAGE env in saas, then it will take config from ci-manager config itself. That's how we handle versioning for OOTB step images e.g. git clone image. It is taken exclusively from ci-manager config only for saas since GIT_CLONE_IMAGE is not set in saas env.

jamie-harness:
This would bind the release of the addon with CI manager, which might not be wanted with Onprem release. @harshjain12 Can you please take a look?

310-ci-manager/ci-manager-config.yml
```diff
@@ -106,7 +106,7 @@ ciExecutionServiceConfig:
   defaultInternalImageConnector: ''
   pvcDefaultStorageSize: 25600
   delegateServiceEndpointVariableValue: "delegate-service:8080"
-  addonImage: harness/ci-addon:release-84
+  addonImage: harness/ci-addon:release-97
   liteEngineImage: harness/ci-lite-engine:release-38
```
harshjain12, line 110: @shubham149 Lite engine version is correct ?

dockerization/ci-manager/scripts/replace_configs.sh
hemanthmantri: Alright. So I will revert the change to this file and let the OPS ticket take care of unsetting env variables in SaaS environments. I hope that'll be sufficient.

310-ci-manager/ci-manager-config.yml
hemanthmantri: Decided to keep the LE version as is. Only the Addon version is only changing. 38 is beyond what prod LE version is (34).

PR #33158, opened 2022-05-14
techdebt: [DEL-3839]: Move CommandTask and related classes to 930-delegate-tasks

johannesHarness:
This PR moves the CommandTask and related classes to 930-delegate-tasks.
The required quite a few changes, which can be grouped into the following major changes:

### Move `Host`
Host currently is extending Base, which we had to break. Since it's not using createdBy/lastUpdatedBy, we were able to omit those fields and directly implement the interfaces that usually Base is implementing. The JIRA ticket has some screenshots on DB queries in prod for hosts collection showing that the two `...By` fields aren't used: https://harness.atlassian.net/browse/DEL-3839

I've also copied the necessary parts of the `Base.onSave()` function over.

### Break apart `ContainerFamily`
Since the enum was also containing application logic which had references we had to break, we separated out the application logic in the enum implementation into its own classes (+ a factory class), and simplified the enum to only have the enum values and no logic.

### Break apart `ContainerTask`'s and its implementation's application logic and introduce DTOs (Kubernetes, ECS)
The bean classes had quite a good amount of application logic, which arguably shouldn't be there in the first part.
Furthermore, they extend Base and actually use the createdBy/lastUpdateBy fields, which means we can't move them.
As such, we've introduced DTO classes for the beans, and moved out the application logic into their own utils classes, which allowed us to move both the DTOs and the actual logic out of 400-rest.

### Break apart `CommandUnitType`
The enum `CommandUnitType` was piggybacking quite a bit of functionality, too.
Since we are unable to move the CommandUnitType due to the extra behavioral part of the enum, we separated out the logic into a separate class `CommandUnitDescriptor` which has a singleton for every CommandUnitType that can be retrieved via `CommandUnitDescriptor.forType(...)`. The `CommandUnitType` now only contains the different types + their name.

### Introduce DTO for `Command`
_Note: There was an existing CommandDTO class for CommandEntity, to avoid confusion, I've renamed the class to CommandEntityDTO. it's not directly registered in morphia / kryo, so that should be fine I hope_

`Command` itself can't be moved, so we introduced a DTO for it.
This required us to find all the places where we sent the `Command` class to the delegate, which seems to be only for the `CommandTask` (Either via parameter object or generic object[] parameter). To find all the usages, we traces all the usages of `TaskType.COMMAND` enum value, hopefully that should be enough?

## Note to DTOs
In order to avoid any requirement for orchestrated delegate + manager release, we reused the same Kryo registration id for the DTO and removed the original one.
However, for that to work the DTO class has to have the exact same name (and if the fields come from a base class, then even the DTO needs to have the fields coming from a base class with the same name.)
That's why the DTOs have the same name, and are found in the software.wings.beans.dto package.

For more details, see: https://harness.atlassian.net/wiki/spaces/DEL/pages/21087028002/DTOs+and+Kryo+Serialization+Issue

---

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`

---
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/33158)


400-rest/src/main/java/software/wings/beans/command/CommandMapper.java
```diff
@@ -0,0 +1,34 @@
...
+                .stream()
+                .map(commandUnit -> commandUnit instanceof Command ? toCommandDTO((Command) commandUnit) : commandUnit)
+                .collect(Collectors.toList()))
+        .commandUnitType(command.getCommandUnitType())
+        .commandExecutionStatus(command.getCommandExecutionStatus())
+        .deploymentType(command.getDeploymentType())
+        .templateVariables(command.getTemplateVariables())
+        .variables(command.getVariables())
+        // getArtifactNeeded() is not exposed and value isn't used at all, so set it to isArtifactNeeded instead.
+        .artifactNeeded(command.isArtifactNeeded())
```
johannesHarness, line 31: @raghuAtWings that should be okay to use isArtifactNeeded, right? i don't see artifactNeeded actually being used.

400-rest/src/main/java/software/wings/beans/command/CommandUnitDescriptor.java
```diff
@@ -1,19 +1,237 @@
...
+import com.google.common.base.Charsets;
+import com.google.common.io.Resources;
+import java.net.URL;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
 
 /**
- * Created by peeyushaggarwal on 6/27/16.
```
johannesHarness, line 15: revert ... how did that get changed in the first place?

400-rest/src/main/java/software/wings/service/impl/HostServiceImpl.java
```diff
@@ -137,7 +137,8 @@ public Host saveHost(Host appHost) {
       }
       return applicationHost;
     } else {
-      return wingsPersistence.saveAndGet(Host.class, appHost);
+      wingsPersistence.save(appHost);
+      return appHost;
```
johannesHarness, line 141: @raghuAtWings is that okay to replace it that way? (Required as saveAndGet needs the entity to extend base)

930-delegate-tasks/src/main/java/software/wings/beans/command/HelmDummyCommandUnit.java
```diff
@@ -17,7 +17,15 @@
 @OwnedBy(CDP)
 @TargetModule(HarnessModule._957_CG_BEANS)
 public class HelmDummyCommandUnit extends AbstractCommandUnit {
+  /**
+   * Required for CommandUnitDescriptor.newInstance
+   */
+  public HelmDummyCommandUnit() {
+    super(CommandUnitType.HELM_DUMMY);
```
johannesHarness, line 24: @raghuAtWings the type for helmdummy was null before - so I added it.

930-delegate-tasks/src/main/java/software/wings/yaml/templatelibrary/TemplateVariableYaml.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.Data;
+
+/**
+ * @author abhinav
+ */
+
+@Data
+@Builder
+public class TemplateVariableYaml {
+  String name;
```
johannesHarness, line 20: make private

400-rest/src/main/java/software/wings/beans/command/CommandMapper.java
johannesHarness: Discussed offline.

johannesHarness:
trigger ti0

930-delegate-tasks/src/main/java/software/wings/beans/command/HelmDummyCommandUnit.java
johannesHarness: Discussed offline - should be good.

400-rest/src/main/java/software/wings/service/impl/HostServiceImpl.java
johannesHarness: `saveAndGet` specifically says:

> @Deprecated
  // There is no point of using this method. Instead use save, and your object will be updated.
  <T extends Base> T saveAndGet(Class<T> cls, T t);

The `Host.onSave()` method also updates the object itself.

raghuAtWings:
*[`400-rest/src/main/java/software/wings/beans/container/ContainerTask.java` line 106 at r2](https://reviewable.io/reviews/harness/harness-core/33158#-N2HvODyDBu046FiWCsN:-N2HvODyDBu046FiWCsO:bzon86) ([raw file](https://github.com/harness/harness-core/blob/bf3cc4532dc31a4997a99722bd14250870044d7e/400-rest/src/main/java/software/wings/beans/container/ContainerTask.java#L106)):*
> ```Java
>   }
> 
>   public static Pattern compileRegexPattern(String domainName) {
> ```

where is this moved?

raghuAtWings:
*[`400-rest/src/main/java/software/wings/beans/container/EcsContainerTask.java` line 226 at r2](https://reviewable.io/reviews/harness/harness-core/33158#-N2Hviv9C_Sua0GTF6AF:-N2Hviv9C_Sua0GTF6AG:bfdkp2j) ([raw file](https://github.com/harness/harness-core/blob/bf3cc4532dc31a4997a99722bd14250870044d7e/400-rest/src/main/java/software/wings/beans/container/EcsContainerTask.java#L226)):*
> ```Java
>   }
> 
>   public TaskDefinition createTaskDefinition(
> ```

where do these go? is it deleted as not used or moved somewhere?

raghuAtWings:
*[`980-commons/src/main/java/io/harness/serializer/HKryo.java` line 107 at r2](https://reviewable.io/reviews/harness/harness-core/33158#-N2HxhtQCKC4JxSZH3se:-N2HxhtQCKC4JxSZH3sf:b-j7ho24) ([raw file](https://github.com/harness/harness-core/blob/aa96e9717f34fecdf2a03fc10360788c55e1eae6/980-commons/src/main/java/io/harness/serializer/HKryo.java#L107)):*
> ```Java
>    * Creates a new HKryo instance.
>    * @param classResolver the class resolver.
>    * @param skipHarnessClassOriginRegistrarCheck if true, classes can be registered by registrars from other sources.
> ```

lets make comment here that its for testing only and should not be used in production code

raghuAtWings:
*[`980-commons/src/main/java/io/harness/serializer/KryoSerializer.java` line 64 at r2](https://reviewable.io/reviews/harness/harness-core/33158#-N2Hxxcn5cz502LridLe:-N2Hxxcn5cz502LridLf:b-2guby3) ([raw file](https://github.com/harness/harness-core/blob/aa96e9717f34fecdf2a03fc10360788c55e1eae6/980-commons/src/main/java/io/harness/serializer/KryoSerializer.java#L64)):*
> ```Java
>    * Creates a new kryo serializer.
>    * @param registrars the set of registrars
>    * @param skipHarnessClassOriginRegistrarCheck if true, classes can be registered by registrars from other sources.
> ```

same here

raghuAtWings (approved):
> Reviewed 170 of 211 files at r1, 51 of 51 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/33158)* status: :shipit: complete! ✓

PR #33177, opened 2022-05-16
fix: [DEL-4086]: Change Kryo CachedFieldNameStrategy to DEFAULT

johannesHarness:
As part of removing 400-rest dependency from 260-delegate we are replacing some storage entities that are sent to the delegate with DTO classes that have only the necessary fields.
To avoid constant delegate / manager release orchestrations for every DTO required when using a new KRYO registrationId for the DTO, we are reusing the same one.

Theoretically, this should be compatible between any manager and delegate version, as we still send all required fields.
However, we noticed that KRYO is using the classname in the field serialization, which means that for example for a class `Original` with field `name` it will serialize it as `Original.name`. Because of that, it is unable to find that same field in `OriginalDto` during deserialization (and the other way around).

This is caused by the following line in our HKryo class:
`getFieldSerializerConfig().setCachedFieldNameStrategy(FieldSerializer.CachedFieldNameStrategy.EXTENDED);`

This PR is changing the strategy to DEFAULT, which won't include the class name in the serialized fields and thus allows seemless serialization between the original and the DTO.

Also adding a UT to verify the behavior and ensure it isn't changed!

## IMPORTANT
This has the potential impact of breaking behavior in case we have classes where the BASE class fields with the same name as the extending class - though that hopefully shouldn't be the case.

Also, we are trying to see if there are any other implications.

---- 

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`

---
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/33177)

### DIFF
--- a/980-commons/BUILD.bazel
+++ b/980-commons/BUILD.bazel
@@ -113,6 +113,7 @@ java_library(
         "//990-commons-test:module",
         "//999-annotations:module",
         "@maven//:com_auth0_java_jwt",
+        "@maven//:com_esotericsoftware_kryo",
         "@maven//:com_fasterxml_jackson_core_jackson_annotations",
         "@maven//:com_fasterxml_jackson_core_jackson_core",
         "@maven//:com_fasterxml_jackson_core_jackson_databind",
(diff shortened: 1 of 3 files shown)


johannesHarness:
trigger messagecheck

PR #33519, opened 2022-05-24
feat: [DEL-4118]: JDK11 watcher fresh install

mdmilic:
This change removes watch & delegate version check from the startup script. With latest watcher version, new startup script will be downloaded from manager and replace this one, so this one is only used for running watcher on fresh delegate pod. 
Delegate version check was not really needed here since watcher does that anyway. Watcher version check was used to always enforce the exact watcher version configured by manager. I think this was a bad idea and should be relaxed making the startup process simpler (initial `start.sh` is coupled with built in `watcher.jar` which is better since it is used to start it). 
This change means that on a fresh container, watcher will always start with a version that is built in, check with manager if it needs upgrade and follow its standard upgrade procedure if so. This is also better because upgrade procedure is in one place fewer now. To avoid running some really old watcher version we should make sure to rebuild delegate images whenever we publish new watcher version (baking it into the image as well). 

Please read [go/trigger-cmd](https://harness.atlassian.net/wiki/spaces/BT/pages/21082441647/PR+Checks) to re-trigger PR Checks using Github comments.

### DIFF
--- a/dockerization/delegate/scripts/start.sh
+++ b/dockerization/delegate/scripts/start.sh
@@ -4,36 +4,6 @@
 # that can be found in the licenses directory at the root of this repository, also available at
 # https://polyformproject.org/wp-content/uploads/2020/05/PolyForm-Free-Trial-1.0.0.txt.
 
-function jar_app_version() {
-  JAR=$1
-  if unzip -l $JAR | grep -q io/harness/versionInfo.yaml
-  then
-    VERSION=$(unzip -c $JAR io/harness/versionInfo.yaml | grep "^version " | cut -d ":" -f2 | tr -d " " | tr -d "\r" | tr -d "\n")
-  fi
-
-  if [ -z "$VERSION" ]
-  then
-    if unzip -l $JAR | grep -q main/resources-filtered/versionInfo.yaml
-    then
-      VERSION=$(unzip -c $JAR main/resources-filtered/versionInfo.yaml | grep "^version " | cut -d ":" -f2 | tr -d " " | tr -d "\r" | tr -d "\n")
-    fi
-  fi
-
-  if [ -z "$VERSION" ]
-  then
-    if unzip -l $JAR | grep -q BOOT-INF/classes/main/resources-filtered/versionInfo.yaml
-    then
-      VERSION=$(unzip -c $JAR BOOT-INF/classes/main/resources-filtered/versionInfo.yaml | grep "^version " | cut -d ":" -f2 | tr -d " " | tr -d "\r" | tr -d "\n")
-    fi
-  fi
-
-  if [ -z "$VERSION" ]
-  then
-    VERSION=$(unzip -c $JAR META-INF/MANIFEST.MF | grep Application-Version | cut -d "=" -f2 | tr -d " " | tr -d "\r" | tr -d "\n")
-  fi
-  echo $VERSION
-}
-
 # url-encodes a given input string - used to encode the proxy password for curl commands.
 # Note:
 #   - We implement the functionality ourselves to avoid dependencies on new packages.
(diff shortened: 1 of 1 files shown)


gaurav-jain1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/33519)* status: 1 file(s) to review. 1 change requests. 0 of 1 approvals obtained. 

___
*[`dockerization/delegate/scripts/start.sh` line 1 at r1](https://reviewable.io/reviews/harness/harness-core/33519#-N2tpnaND9W4aa69gPc2:-N2tpnaND9W4aa69gPc3:b-kk2jwg) ([raw file](https://github.com/harness/harness-core/blob/e5a36d5f34e1771d0cacfb7cf111efed6719d952/dockerization/delegate/scripts/start.sh#L1)):*
> ```Shell
> #!/bin/bash -e
> ```

We can also change start.sh.ftl file to remove version check



_Code quote:_
```Shell
#!/bin/bash -e
```
___
*[`dockerization/delegate/scripts/start.sh` line 167 at r1](https://reviewable.io/reviews/harness/harness-core/33519#-N2tpi7G3G4XYUcK9GXU:-N2tpi7G3G4XYUcK9GXV:b-rzopmc) ([raw file](https://github.com/harness/harness-core/blob/e5a36d5f34e1771d0cacfb7cf111efed6719d952/dockerization/delegate/scripts/start.sh#L167)):*
> ```Shell
> 
> if [ ! -d $JRE_DIR -o ! -e $JRE_BINARY ]; then
>   echo "Downloading JRE packages..."
> ```

Do we don't want to download jre anymore ?



_Code quote:_
```Shell
if [ ! -d $JRE_DIR -o ! -e $JRE_BINARY ]; then
  echo "Downloading JRE packages..."
  JVM_TAR_FILENAME=$(basename "$JVM_URL")
  curl $MANAGER_PROXY_CURL -#kLO $JVM_URL
  echo "Extracting JRE packages..."
  rm -rf $JRE_DIR
  tar xzf $JVM_TAR_FILENAME
  rm -f $JVM_TAR_FILENAME
fi
```

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/33519)* status: 1 file(s) to review. 1 change requests. 2. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`dockerization/delegate/scripts/start.sh` line 1 at r1](https://reviewable.io/reviews/harness/harness-core/33519#-N2tpnaND9W4aa69gPc2:-N2uRsn05_IeHsOzVGG8:bmys2de) ([raw file](https://github.com/harness/harness-core/blob/e5a36d5f34e1771d0cacfb7cf111efed6719d952/dockerization/delegate/scripts/start.sh#L1)):*

Yes, but I would like to do it in a separate change (after we rollout JDK11). This one will unblock JDK11 watcher rollout and is smaller, less impact
</blockquote></details>

Also, that would make it manager change, now it's only docker image change

gaurav-jain1 (approved):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/33519)* status: 1 file(s) to review. 

___
*[`dockerization/delegate/scripts/start.sh` line 1 at r1](https://reviewable.io/reviews/harness/harness-core/33519#-N2tpnaND9W4aa69gPc2:-N2vHO8DBEtRYXdtpoXj:bjlrmyw) ([raw file](https://github.com/harness/harness-core/blob/e5a36d5f34e1771d0cacfb7cf111efed6719d952/dockerization/delegate/scripts/start.sh#L1)):*

Also, that would make it manager change, now it's only docker image change
</blockquote></details>

ok, although we can merge it to develop

PR #34679, opened 2022-06-24
feat: [CI-4692]: Git clone step implementation

ricksjames:
## Checklist

## Comment Triggers
<details>
  
</details>

<details>
  
- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
</details>

---
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/34679)


ricksjames:
Note this is still being worked and not ready for review yet. (also needs rebased)

shubham149 (commented):
Have you tested it on both k8 and vm? 

I see a couple of issues with this new step.
1. GitCloneStepInfo needs to implement PluginCompatibleStep. Check EcrStepInfo
2. There will be collision of environment variables of inbuilt git clone step and explicit step i.e. some attributes of inbuilt git clone step will also be set on this explicit step.
3. Have we got ack from PM for the git clone step yaml? yaml seems odd to me e.g. inclusion of pr clone strategy in it.
4. Test it for all the possible cases & add automation for it in both vm (linux & windows) and k8 (linux & windows)

shubham149:
Basic understanding of how plugin git clone will be required for this task https://github.com/wings-software/drone-git

930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/SecretSpecBuilder.java
```diff
@@ -327,8 +327,10 @@ private Map<String, SecretParams> retrieveGitHubSecretParams(
         if (isEmpty(username)) {
           throw new CIStageExecutionException("Github connector should have not empty username");
         }
+        String connectorIdentifier = gitConnector.getIdentifier();
         secretData.put(DRONE_NETRC_USERNAME,
-            SecretParams.builder().secretKey(DRONE_NETRC_USERNAME).value(encodeBase64(username)).type(TEXT).build());
+            SecretParams.builder().secretKey(DRONE_NETRC_USERNAME + "_" + connectorIdentifier)
```
harshjain12, line 332: we should not use connector identifier

Connector identifier can be same for account and project level

320-ci-execution/src/main/java/io/harness/ci/buildstate/CodebaseUtils.java
```diff
@@ -164,6 +164,15 @@ public Map<String, String> getRuntimeCodebaseVars(Ambiance ambiance) {
   }
 
   public Map<String, String> getGitEnvVariables(ConnectorDetails gitConnector, CodeBase ciCodebase) {
+    if (ciCodebase == null) {
+      throw new IllegalArgumentException("CI codebase spec is not set");
```
harshjain12, line 168: throw CIStageExecutionException

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeStepUtils.java
```diff
@@ -918,4 +979,103 @@ private ContainerResource getContainerResource(StepElementConfig stepElement) {
...
+            .settings(ParameterField.createValueField(settings))
+            .envVariables(envVariables)
+            .entrypoint(entrypoint)
+            .harnessManagedImage(true)
+            .resources(gitCloneStepInfo.getResources())
+            .privileged(ParameterField.createValueField(null))
+            .reports(ParameterField.createValueField(null))
+            .build();
+
+    if (accountId != null) {
```
harshjain12, line 1041: Use isNotEmpty

What if accountId is empty ?
Although it should not be empty

harshjain12 (changes requested):
> Reviewed 1 of 40 files at r1.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/34679)* status: 39 file(s) to review. 1 change requests. 0 of 1 approvals obtained.

ricksjames:
> Have you tested it on both k8 and vm?
> 
> I see a couple of issues with this new step.
> 
>     1. GitCloneStepInfo needs to implement PluginCompatibleStep. Check EcrStepInfo
> 
>     2. There will be collision of environment variables of inbuilt git clone step and explicit step i.e. some attributes of inbuilt git clone step will also be set on this explicit step.
> 
>     3. Have we got ack from PM for the git clone step yaml? yaml seems odd to me e.g. inclusion of pr clone strategy in it.
> 
>     4. Test it for all the possible cases & add automation for both vm (linux & windows) and k8 (linux & windows)

@shubham149 with my latest force push I've addressed the following:

1. fixed
2. I have specifically handled environment variable conflicts for:
DRONE_COMMIT_BRANCH
DRONE_TAG
DRONE_REMOTE_URL
DRONE_NETRC_MACHINE
DRONE_NETRC_USERNAME
DRONE_NETRC_PASSWORD
and will fix any more conflicts that I uncover
3. Removed prCloneStrategy
4. Right now I have only been focusing on k8 linux and github. Will test other combinations as I develop them. I have a separate task for automation testing also.

930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/SecretSpecBuilder.java
ricksjames: Fixed. I added  connectorIdentifier + projectIdentifier +  orgIdentifier for the key name. As far as I can tell this should be unique. Let me know otherwise.

320-ci-execution/src/main/java/io/harness/ci/buildstate/CodebaseUtils.java
ricksjames: fixed.

930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/SecretSpecBuilder.java
```diff
@@ -328,9 +328,16 @@ private Map<String, SecretParams> retrieveGitHubSecretParams(
...
           throw new CIStageExecutionException("Github connector should have not empty username");
         }
-        String connectorIdentifier = gitConnector.getIdentifier();
+        StringBuilder uniqueConnectorIdentifier = new StringBuilder();
+        uniqueConnectorIdentifier.append("_").append(gitConnector.getIdentifier());
+        if(isNotEmpty(gitConnector.getOrgIdentifier())) {
+          uniqueConnectorIdentifier.append("_").append(gitConnector.getOrgIdentifier());
+        }
+        if(isNotEmpty(gitConnector.getProjectIdentifier())) {
+          uniqueConnectorIdentifier.append("_").append(gitConnector.getProjectIdentifier());
```
harshjain12, line 337: Think about character limits and special character in getOrgIdentifier and getProjectIdentifier

harshjain12 (changes requested):

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeStepUtils.java
ricksjames: Fixed. Added some TODOs for the null case. I overloaded this method for VM code path use (which is why the null path case was needed), but that may not work anyway, so when I get to testing the VM path I'll see if I need to make a new implementation.

930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/SecretSpecBuilder.java
ricksjames: fixed.

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeStepUtils.java
```diff
@@ -488,6 +513,41 @@ private ContainerDefinitionInfo createRunTestsStepContainerDefinition(RunTestsSt
...
+
+    //Set the Git Connector Reference environment variables
+    Map<String, String> gitEnvVars = codebaseUtils.getGitEnvVariables(gitConnector,
+            gitCloneStepInfo.getRepoName().getValue());
+    pluginStepInfo.getEnvVariables().putAll(gitEnvVars);
+
+    String cloneDirectoryString = RunTimeInputHandler.resolveStringParameter("cloneDirectory", name,
+            identifier, gitCloneStepInfo.getCloneDirectory(), false);
+    if(isNotEmpty(cloneDirectoryString)) {
+      pluginStepInfo.getEnvVariables().put(HARNESS_WORKSPACE, cloneDirectoryString);
```
harshjain12, line 537: This could be an issue.
Can you check if log upload is working fine. 
We also will have to check default permissions given to this directory. It should be same as that of default harness workspace.

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeStepUtils.java
```diff
@@ -488,6 +513,41 @@ private ContainerDefinitionInfo createRunTestsStepContainerDefinition(RunTestsSt
...
+    }
+
+    //TODO: not sure if this is the best way to do this. Would GitCloneStep ever need the other information from ExecutionSource?
+    //Don't use the ciExecutionArgs ExecutionSource (contains codebase build info that overwrites branch/tag env variables
+    ciExecutionArgs.setExecutionSource(null);
```
harshjain12, line 542: do not update ciExecutionArgs

320-ci-execution/src/main/java/io/harness/ci/serializer/vm/VmStepSerializer.java
```diff
@@ -47,6 +52,13 @@ public VmStepInfo serialize(Ambiance ambiance, CIStepInfo stepInfo, StageInfraDe
       case RUN_TESTS:
         return vmRunTestStepSerializer.serialize(
             (RunTestsStepInfo) stepInfo, identifier, parameterFieldTimeout, stepName, ambiance);
+      case GIT_CLONE:
+        //TODO: this has not been tested yet and likely doesn't work
```
harshjain12, line 56: can you check todos

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeTaskParamsBuilder.java
```diff
@@ -290,15 +295,24 @@ private void consumePortDetails(Ambiance ambiance, List<ContainerDefinitionInfo>
...
+      Map<String, String> runtimeCodebaseVars,
+      Map<String, List<K8BuildJobEnvInfo.ConnectorConversionInfo>> connectorRefs, ConnectorDetails gitConnector,
+      Map<String, String> volumeToMountPath, ContainerSecurityContext ctrSecurityContext,
+      String logPrefix, List<SecretVariableDetails> secretVariableDetails,
+      Map<String, ConnectorDetails> githubApiTokenFunctorConnectors, OSType os) {
     Map<String, String> envVars = new HashMap<>();
+
+    String workDirPath = k8InitializeTaskUtils.getWorkDir();
+    //set default workspace before containerDefinitionInfo so default can be overwritten
+    envVars.put(HARNESS_WORKSPACE, workDirPath);
```
harshjain12, line 308: why do we have to add this statement ?
Below condition should take care of it.

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeStepUtils.java
```diff
@@ -937,4 +1000,118 @@ private ContainerResource getContainerResource(StepElementConfig stepElement) {
...
+    return step;
+  }
+
+  /**
+   * Get Build Env variable - branch or tag
+   *
+   * @param gitCloneStepInfo gitCloneStepInfo
+   * @return a pair containing whether the build is configured for a branch or tag, and the value of the branch or tag
+   */
+  private static Pair<String, String> getBuildEnvVar(GitCloneStepInfo gitCloneStepInfo) {
```
harshjain12, line 1085: this looks like a copy of io.harness.ci.integrationstage.IntegrationStageUtils#treatWebhookAsManualExecution

harshjain12: Is there a way to club both the methods in single ?

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeStepUtils.java
```diff
@@ -488,6 +513,41 @@ private ContainerDefinitionInfo createRunTestsStepContainerDefinition(RunTestsSt
         .build();
   }
 
+  private ContainerDefinitionInfo createGitCloneStepContainerDefinition(GitCloneStepInfo gitCloneStepInfo,
```
harshjain12, line 516: UTs for this method is important

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeTaskParamsBuilder.java
ricksjames: Originally the workspace env var was set in K8InitializeTaskUtils.getCommonStepEnvVariables. I removed it from being added there because these commonEnvVars are added last to the container (ln362) to override all others which is now not the desired behavior. 
Line 306 is getting the default (which is no longer being set in commonEnvVars).
Line 308 is setting the env var to that default
Line 311 gets the container definition info, which in the case of the git clone step, can set the HARNESS_WORKSPACE, writing over the default in the map 
line 312 gets either the default, or the overridden value from the map for further use in the method

320-ci-execution/src/main/java/io/harness/ci/serializer/vm/VmStepSerializer.java
ricksjames: implemented and tested functionality

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeStepUtils.java
ricksjames: Took a pass at this, but treatWebhookAsManualExecution has pretty specific logic, so I was only able to pull out some of the logic.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/InitializeStepUtils.java
```diff
@@ -0,0 +1,156 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+
+public class InitializeStepUtils {
```
shubham149, line 40: Are we using this code?

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeStepUtils.java
```diff
@@ -325,6 +329,10 @@ private ContainerDefinitionInfo createStepContainerDefinition(StepElementConfig
         return createPluginCompatibleStepContainerDefinition((PluginCompatibleStep) ciStepInfo, integrationStage,
             ciExecutionArgs, portFinder, stepIndex, stepElement.getIdentifier(), stepElement.getName(),
             stepElement.getType(), timeout, accountId, os, extraMemoryPerStep, extraCPUPerStep);
+      case GIT_CLONE:
```
shubham149, line 332: git clone functionality needs to go inside pluginCompatibleStep functionality

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeStepUtils.java
```diff
@@ -509,6 +517,41 @@ private ContainerDefinitionInfo createRunTestsStepContainerDefinition(RunTestsSt
         .build();
   }
 
+  private ContainerDefinitionInfo createGitCloneStepContainerDefinition(GitCloneStepInfo gitCloneStepInfo,
```
shubham149, line 520: This should be a part of pluginCompatibleStep

diptiman-adak:
 


            
 #### 11 Issues 
* Vulnerability 2 Vulnerabilities 
 * Code issue 9 Code issues 

 [Details](https://app.embold.io/organization/gh/harness/repositories/7752dbb3099d2dba3b5d37f5931a5e4c/pull-request/34679)

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/InitializeStepUtils.java
harshjain12, line 40: make it singleton

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/IntegrationStageUtils.java
```diff
@@ -198,12 +201,14 @@ public static boolean treatWebhookAsManualExecution(
     String url = getGitURLFromConnector(connectorDetails, codeBase);
     WebhookExecutionSource webhookExecutionSource = WebhookTriggerProcessorUtils.convertWebhookResponse(parsedPayload);
     Build build = RunTimeInputHandler.resolveBuild(codeBase.getBuild());
-    if (build != null) {
-      if (build.getType() == BuildType.PR) {
-        ParameterField<String> number = ((PRBuildSpec) build.getSpec()).getNumber();
-        String numberString =
-            RunTimeInputHandler.resolveStringParameter("number", "Git Clone", "identifier", number, false);
-        if (!numberString.equals(PR_EXPRESSION)) {
+    Pair<BuildType, String> buildTypeAndValue = getBuildTypeAndValue(build);
```
harshjain12, line 204: Why do we have to change this method ?

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeTaskParamsBuilder.java
ricksjames: OBE, now using DRONE_WORKSPACE

332-ci-manager/service/src/main/java/io/harness/ci/execution/states/AbstractStepExecutable.java
```diff
@@ -318,6 +324,14 @@ private CIExecuteStepTaskParams getVmTaskParams(Ambiance ambiance, VmStepInfo vm
       workingDir = infraDetails.getWorkDir();
     }
 
+    //GitClone step can override the workspace, get and set the value from the env variables
+    if (vmStepInfo instanceof VmPluginStep) {
```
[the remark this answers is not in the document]
ricksjames: Updated to use DRONE_WORKSPACE.
Lets discuss together with Harsh in Aug 4 meeting.

332-ci-manager/service/src/main/java/io/harness/ci/beans/environment/pod/container/ContainerDefinitionInfo.java
```diff
@@ -45,4 +46,5 @@ public class ContainerDefinitionInfo {
   private Boolean privileged;
   private Integer runAsUser;
   private String imagePullPolicy;
+  private ConnectorDetails gitConnector;
```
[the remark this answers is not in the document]
ricksjames: Lets discuss together with Harsh in Aug 4 meeting.

950-delegate-tasks-beans/src/main/java/io/harness/delegate/beans/ci/pod/ContainerParams.java
```diff
@@ -38,6 +38,7 @@ public abstract class ContainerParams implements NestedAnnotationResolver {
   @Deprecated private boolean privileged;
   private String imagePullPolicy;
   private ContainerSecurityContext securityContext;
+  private ConnectorDetails gitConnector;
```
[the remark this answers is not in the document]
ricksjames: Let's discuss together with Harsh in Aug 4 meeting.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
```diff
@@ -290,15 +295,24 @@ private void consumePortDetails(Ambiance ambiance, List<ContainerDefinitionInfo>
...
+
+    String workDirPath = k8InitializeTaskUtils.getWorkDir();
+    //set default workspace before containerDefinitionInfo so default can be overwritten
+    envVars.put(HARNESS_WORKSPACE, workDirPath);
+
     if (isNotEmpty(containerDefinitionInfo.getEnvVars())) {
       envVars.putAll(containerDefinitionInfo.getEnvVars()); // Put customer input env variables
     }
+
+    workDirPath = envVars.get(HARNESS_WORKSPACE);
```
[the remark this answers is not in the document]
ricksjames: fixed

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
```diff
@@ -335,6 +349,16 @@ private CIK8ContainerParams createCIK8ContainerParams(NGAccess ngAccess,
         k8InitializeTaskUtils.getSecretVariableDetails(ngAccess, containerDefinitionInfo, secretVariableDetails);
 
     Map<String, String> envVarsWithSecretRef = k8InitializeTaskUtils.removeEnvVarsWithSecretRef(envVars);
+    if(GIT_CLONE_STEP_ID.equals(containerDefinitionInfo.getStepIdentifier())) {
```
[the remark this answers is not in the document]
ricksjames: Let's discuss together with Harsh in Aug 4 meeting.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
```diff
@@ -349,12 +373,17 @@ private CIK8ContainerParams createCIK8ContainerParams(NGAccess ngAccess,
     }
     boolean privileged = containerDefinitionInfo.getPrivileged() != null && containerDefinitionInfo.getPrivileged();
 
+    if(!STEP_MOUNT_PATH.equals(workDirPath)) {
```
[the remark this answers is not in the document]
ricksjames: Let's discuss together with Harsh in Aug 4 meeting.
Note this section has been updated since this comment was made, but still applies.
While I'm not positive what is happening under the covers, I'm assuming when I don't add an extra volume to mount for every gitCloneStep cloneDirectory, that directory doesn't get created and I then cannot use it, which likely would be resolved by creating the DRONE_WORKSPACE in the plugin docker image.

332-ci-manager/service/src/main/java/io/harness/ci/beans/steps/stepinfo/GitCloneStepInfo.java
```diff
@@ -0,0 +1,165 @@
...
+  @VariableExpression(skipVariableExpression = true)
+  @Min(MIN_RETRY)
+  @Max(MAX_RETRY)
+  private int retry;
+
+  @NotNull
+  @ApiModelProperty(dataType = STRING_CLASSPATH)
+  ParameterField<String> connectorRef;
+
+  ContainerResource resources;
```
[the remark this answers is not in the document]
ricksjames: Is this comment referring to "ContainerResource resources;"?
I checked all the other *StepInfo types, and I couldn't find any that had wrapped the resources in a ParameterField. Just wanted to double check as the comment doesn't align with the pattern I followed from the other steps.

332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/CodebaseUtils.java
```diff
@@ -164,7 +164,19 @@ public Map<String, String> getRuntimeCodebaseVars(Ambiance ambiance) {
     return codebaseRuntimeVars;
   }
 
-  public Map<String, String> getGitEnvVariables(ConnectorDetails gitConnector, CodeBase ciCodebase) {
+  public Map<String, String> getGitEnvVariables(ConnectorDetails gitConnector, CodeBase ciCodebase,
```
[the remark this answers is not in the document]
ricksjames: getGitEnvVariables(ConnectorDetails gitConnector, CodeBase ciCodebase, boolean skipGitClone)
overloads
getGitEnvVariables(ConnectorDetails gitConnector, String repoName)

Method overloading is a pretty widely used java feature. In this case it is also preventing code duplication as the overloaded method calls the original, and the purpose of the methods are identical. 
What would you prefer instead or what isn't clear?

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeStepUtils.java
ricksjames: fixed

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/InitializeStepUtils.java
ricksjames: yes, 
VmInitializeStepUtils and K8InitializeStepUtils exist in this same package. These utils are used by both K8 and VM, hence I made this new class for them.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeStepUtils.java
ricksjames: GIT_CLONE has unique logic that differs from PluginCompatibleStep functionality. This is where the bulk of that logic is called in the new method createGitCloneStepContainerDefinition(). Note that inside of createGitCloneStepContainerDefinition() I am converting gitCloneStepInfo to PluginStepInfo, and then using the existing createPluginStepContainerDefinition to reuse the plugin logic that already exists there for cloneCodebase.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeStepUtils.java
ricksjames: See above response, in addition, the original cloneCodebase is a pluginStep and used the createPluginStepContainerDefinition() method, not the createPluginCompatibleStep... method. Hence overlaps more with that method (which I call inside createGitCloneStepContainerDefinition() for the overlap)

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/IntegrationStageUtils.java
ricksjames: We don't have to modify it. I made this change to reuse logic from this method in InitializeStepUtils.getBuildEnvVar() as suggested (the reused logic was pulled out into this method - "getBuildTypeAndValue(Build build)" in this class, but in the end the value gained from the reuse isn't significant, and we'd probably be better off leaving this method as is.
Will undo this change.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskUtils.java
```diff
@@ -537,11 +536,7 @@ public Map<String, String> getCommonStepEnvVariables(K8PodDetails k8PodDetails,
     final String stageID = k8PodDetails.getStageID();
     final String executionID = ambiance.getPlanExecutionId();
 
-    // Add git connector environment variables
-    envVars.putAll(gitEnvVars);
```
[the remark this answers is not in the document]
ricksjames: This logic was moved to:
K8InitializeTaskParamsBuilder ln346. It was moved here because it is only used by the codebase plugin step, and the gitEnvVars are not needed in every step (plus they conflict with the new git clone step, as codebase and the gitclonestep can have different configurations).

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskUtils.java
```diff
@@ -537,11 +536,7 @@ public Map<String, String> getCommonStepEnvVariables(K8PodDetails k8PodDetails,
...
 
-    // Add runtime git vars, i.e. manual pull request execution data.
-    envVars.putAll(runtimeCodebaseVars);
```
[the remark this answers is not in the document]
ricksjames: This logic was moved to:
K8InitializeTaskParamsBuilder ln353. It was moved here because it is only used by the codebase plugin step, and the runtimeCodebaseVars are not needed in every step (plus they conflict with the new git clone step, as codebase and the gitclonestep can have different configurations).

332-ci-manager/service/src/main/java/io/harness/ci/beans/environment/pod/container/ContainerDefinitionInfo.java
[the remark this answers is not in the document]
ricksjames: Reviewed with Harsh and Shubham, came to the conclusion that we can remove this and add to the secret env variables, etc using PluginSettingUtils.getConnectorSecretEnvMap() and adding a new method for Git clone step for ConnectorEnvVariablesHelper.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeStepUtils.java
ricksjames: Reviewed with Harsh and Shubham.
There is an alternate pattern established in the code that could also be used. 
Instead of using the case statement inside createStepContainerDefinition, in createPluginCompatibleStepContainerDefinition the following method is called:
PluginSettingUtils.getPluginCompatibleEnvVariables, where there is another switch statement for PluginCompatibleSteps. 
Harsh and I to review further. There may be some benefit for code clarity to move Git_Clone logic here, but considering cloneCodebase uses PluginStep, there may actually be more overlap and less code duplication by leaving code as is.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskUtils.java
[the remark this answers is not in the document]
ricksjames: Discussed with Harsh and Shubham, 
Will fix.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskUtils.java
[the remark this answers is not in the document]
ricksjames: Discussed with Harsh and Shubham, 
Will fix.

930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/CIK8InitializeTaskHandler.java
```diff
@@ -504,13 +503,10 @@ private Map<String, String> getAndUpdateDelegateServiceToken(CIK8ContainerParams
   }
 
   private Map<String, String> getAndUpdateGitSecretData(
```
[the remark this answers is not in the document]
ricksjames: will update.

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeStepUtils.java
ricksjames: This logic was moved to PluginSettingUtils, and tests were created in PluginSettingUtilsTest.
Done.

332-ci-manager/service/src/main/java/io/harness/ci/beans/steps/stepinfo/GitCloneStepInfo.java
```diff
@@ -0,0 +1,165 @@
...
+  @YamlSchemaTypes({runtime})
+  @ApiModelProperty(dataType = SwaggerConstants.INTEGER_CLASSPATH)
+  ParameterField<Integer> depth;
+
+  @YamlSchemaTypes({runtime})
+  @ApiModelProperty(dataType = SwaggerConstants.BOOLEAN_CLASSPATH)
+  ParameterField<Boolean> sslVerify;
+
+  @ApiModelProperty(dataType = STRING_CLASSPATH)
+  ParameterField<String> cloneDirectory;
```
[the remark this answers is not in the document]
ricksjames: In the UI the provided value is /harness. In PluginSettingUtils this is used as an optional field. Meaning if it is not provided, DRONE_WORKSPACE will not be set, and the docker image will default to the HARNESS_WORKSPACE of /harness.

332-ci-manager/service/src/main/java/io/harness/ci/beans/environment/pod/container/ContainerDefinitionInfo.java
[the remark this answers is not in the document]
ricksjames: fixed

332-ci-manager/service/src/main/java/io/harness/ci/beans/steps/stepinfo/GitCloneStepInfo.java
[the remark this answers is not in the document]
ricksjames: @shubham149 can you provide further clarification for above?

950-delegate-tasks-beans/src/main/java/io/harness/delegate/beans/ci/pod/ContainerParams.java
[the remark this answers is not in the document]
ricksjames: fixed

933-ci-commons/src/main/java/io/harness/connector/SecretSpecBuilder.java
```diff
@@ -154,17 +159,19 @@ public Map<String, SecretParams> decryptConnectorSecretVariables(Map<String, Con
   }
 
   public Map<String, SecretParams> decryptConnectorSecret(ConnectorDetails connectorDetails) {
-    log.info("Decrypting connector id:[{}], type:[{}]", connectorDetails.getIdentifier(),
-        connectorDetails.getConnectorType());
+    ConnectorType type = connectorDetails.getConnectorType();
```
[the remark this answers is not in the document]
ricksjames: fixed

930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/CIK8InitializeTaskHandler.java
```diff
@@ -120,7 +120,6 @@ public K8sTaskExecutionResponse executeTaskInternal(
     Stopwatch timer = Stopwatch.createStarted();
     CIK8InitializeTaskParams cik8InitializeTaskParams = (CIK8InitializeTaskParams) ciInitializeTaskParams;
     String cik8BuildTaskParamsStr = cik8InitializeTaskParams.toString();
-    ConnectorDetails gitConnectorDetails = cik8InitializeTaskParams.getCik8PodParams().getGitConnector();
```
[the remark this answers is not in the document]
ricksjames: OBE, this line not changed with the new implemenation.

930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/CIK8InitializeTaskHandler.java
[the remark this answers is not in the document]
ricksjames: fixed. DRONE_NETRC_USERNAME and DRONE_NETRC_PASSWORD are available as before in the containers. These will only be replaced for the Git Clone step, where other values are needed by the step.

332-ci-manager/service/src/main/java/io/harness/ci/execution/states/AbstractStepExecutable.java
```diff
@@ -540,6 +554,12 @@ private UnitStep serialiseStep(CIStepInfo ciStepInfo, String taskId, String logK
       case RUN:
         return runStepProtobufSerializer.serializeStepWithStepParameters((RunStepInfo) ciStepInfo, port, taskId, logKey,
             stepIdentifier, ParameterField.createValueField(Timeout.fromString(timeout)), accountId, stepName);
+      case GIT_CLONE:
```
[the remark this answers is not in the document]
ricksjames: fixed

332-ci-manager/service/src/main/java/io/harness/ci/execution/states/AbstractStepExecutable.java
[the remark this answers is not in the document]
ricksjames: fixed

332-ci-manager/service/src/main/java/io/harness/ci/execution/serializer/vm/VmStepSerializer.java
```diff
@@ -47,6 +49,9 @@ public VmStepInfo serialize(Ambiance ambiance, CIStepInfo stepInfo, StageInfraDe
       case RUN_TESTS:
         return vmRunTestStepSerializer.serialize(
             (RunTestsStepInfo) stepInfo, identifier, parameterFieldTimeout, stepName, ambiance);
+      case GIT_CLONE:
+        return vmGitCloneStepSerializer.serialize(
```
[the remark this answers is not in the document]
ricksjames: fixed

332-ci-manager/service/src/main/java/io/harness/ci/execution/serializer/vm/VmGitCloneStepSerializer.java
```diff
@@ -0,0 +1,93 @@
...
+import io.harness.pms.yaml.ParameterField;
+import io.harness.yaml.core.timeout.Timeout;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import java.util.HashMap;
+import java.util.Map;
+
+@Singleton
+public class VmGitCloneStepSerializer {
```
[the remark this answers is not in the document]
ricksjames: Fixed.
This is now managed in the PluginCompatibleStep flow, with logic either already existing in that flow, or defined in PluginSettingUtils.getGitCloneStepInfoEvnVariables().

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskUtils.java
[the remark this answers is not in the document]
ricksjames: fixed.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskUtils.java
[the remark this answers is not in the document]
ricksjames: fixed

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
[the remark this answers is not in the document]
ricksjames: fixed.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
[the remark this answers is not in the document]
ricksjames: fixed.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/InitializeStepUtils.java
ricksjames: Fixed.
This is now managed in the PluginCompatibleStep flow, with logic either already existing in that flow, or defined in PluginSettingUtils.getGitCloneStepInfoEvnVariables().

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeStepUtils.java
ricksjames: After review I determined the PluginCompatible flow suggested was simpler, code has been updated to follow that flow. Fixed.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeStepUtils.java
ricksjames: After review I determined the PluginCompatibleStep flow suggested by Shubham was simpler; code has been updated to follow that flow. Fixed.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/InitializeStepUtils.java
ricksjames: OBE, this class no longer exists.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/IntegrationStageUtils.java
ricksjames: Fixed.

320-ci-execution/src/main/java/io/harness/ci/integrationstage/K8InitializeStepUtils.java
ricksjames: @harshjain12 Do you still suspect an issue here? We are now using the DRONE_WORKSPACE env variable and creating the clone directory and changing to it in the drone-git docker image.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
```diff
@@ -332,6 +338,17 @@ private CIK8ContainerParams createCIK8ContainerParams(NGAccess ngAccess,
         k8InitializeTaskUtils.getSecretVariableDetails(ngAccess, containerDefinitionInfo, secretVariableDetails);
 
     Map<String, String> envVarsWithSecretRef = k8InitializeTaskUtils.removeEnvVarsWithSecretRef(envVars);
+
+    //remove any codebase env vars from commonEnvVars which will overwrite git clone step envVars
+    String droneRemoteUrl = envVars.get(DRONE_REMOTE_URL);
```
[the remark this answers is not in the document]
shubham149: it should be a part of PluginSettingUtils

332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java
```diff
@@ -185,6 +218,8 @@ public static Map<EnvVariableEnum, String> getConnectorSecretEnvMap(CIStepInfoTy
         map.put(EnvVariableEnum.ARTIFACTORY_USERNAME, PLUGIN_USERNAME);
         map.put(EnvVariableEnum.ARTIFACTORY_PASSWORD, PLUGIN_PASSW);
         return map;
+      case GIT_CLONE:
+        return map;
```
shubham149, line 222: Shouldn't we have to add git related secret variables here?

332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java
```diff
@@ -662,6 +697,117 @@ private static Map<String, String> getUploadToS3StepInfoEnvVariables(UploadToS3S
...
+    final NGAccess ngAccess = AmbianceUtils.getNgAccess(ambiance);
+    final ConnectorDetails gitConnector = codebaseUtils.getGitConnector(ngAccess, connectorRef);
+
+    // Set the Git Connector Reference environment variables
+    final String repoName = gitCloneStepInfo.getRepoName().getValue();
+    final Map<String, String> gitEnvVars = codebaseUtils.getGitEnvVariables(gitConnector, repoName);
+    map.putAll(gitEnvVars);
+
+    Pair<String, String> buildEnvVar = getBuildEnvVar(gitCloneStepInfo);
+    if (buildEnvVar != null) {
```
shubham149, line 718: What happens if the value is null?

332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java
```diff
@@ -662,6 +697,117 @@ private static Map<String, String> getUploadToS3StepInfoEnvVariables(UploadToS3S
...
+      String type = buildEnvVar.getKey();
+      setMandatoryEnvironmentVariable(map, type, buildEnvVar.getValue());
+      if (DRONE_TAG.equals(type)) {
+        setMandatoryEnvironmentVariable(map, DRONE_BUILD_EVENT, TAG_BUILD_EVENT);
+      }
+    }
+
+    String cloneDir =
```
shubham149, line 726: What will be the default value of clone directory if it is unset?

shubham149: If 2 git clone steps are present without clone directory set, what will happen?

930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/CIK8InitializeTaskHandler.java
```diff
@@ -337,6 +337,11 @@ private void createEnvVariablesSecrets(CoreV1Api coreV1Api, String namespace,
     List<CIK8ContainerParams> containerParamsList = podParams.getContainerParamsList();
     String k8SecretName = getSecretName(podParams.getName());
 
+    log.info("Creating git secret env variables for pod: {}", podParams.getName());
```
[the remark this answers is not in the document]
ricksjames: The moved block of code sets the container secrets to the values from the **pod** gitConnector (clone codebase connector). With the new location - the clone codebase connector secrets are still added to every container, but now put in the map before the gitCloneStep secrets. This way if the **container** has gitConnector secrets (which is the case for git clone step), they will replace the clone codebase secrets, otherwise, clone codebase will be available.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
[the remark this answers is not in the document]
ricksjames: @shubham149 
Currently PluginSettingUtils.getPluginCompatibleEnvVariables(...) returns a map of env variables. So with the current implementation, this method can only add env variables, not remove them. To enable it to add env variables we would need to pass the env variable map into this method, then we could remove these env variables with PluginSettingUtils code. 

From a Java perspective it is generally preferable to return values from a function instead of modify the value of a passed in parameter for the following reasons:
1. Clearer code - most developers probably don't expect the map to be modified by the method (could be a little more clear if we renamed all the methods to something like - setEnvVariablesToPluginCompatibleEnvVars...)
2. Unit testing is easier when returning values 

Based on that thought process, I determined leaving this logic in K8InitializeTaskParamsBuilder was preferable.

Do you have any other ideas? Or like the idea of passing the env map into the method and then modifying it?

332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java
ricksjames: @shubham149 can you further explain what this section of code does, why it is needed, and what values you would expect me to put here for git clone step? The secret env vars are created on the delegate side without passing any values into this map.

332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java
ricksjames: DRONE_COMMIT_BRANCH, DRONE_TAG, DRONE_BUILD_EVENT will not be set for the container. The clone will fail when the step runs due to missing values.

332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java
ricksjames: When left unset, DRONE_WORKSPACE env var will not be set for the container. When set this env variable causes the drone-git image to create this directory, and cd to it. If unset, the container will remain at the HARNESS_WORKSPACE dir "/harness".

So the clone will occur at /harness. When unset twice, both clones will try to clone to the same directory, and the second one will fail.

ricksjames:
trigger ti2

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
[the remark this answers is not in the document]
ricksjames: Additionally note that commonEnvVars are added last - so they overwrite all other values.

ricksjames:
trigger pmd, ti3

ricksjames:
trigger pmd

shubham149:
*[`332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java` line 718 at r9](https://reviewable.io/reviews/harness/harness-core/34679#-N7dD5kTO_jNhEG74Pui-r9-718:-N9_Phq0BituC6vsVF59:b-5cx0yn) ([raw file](https://github.com/harness/harness-core/blob/d226f796dbd1b6106bcc1b6fb26de6e34c35460d/332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java#L718)):*
<details><summary><i>Previously, ricksjames wrote…</i></summary><blockquote>

DRONE_COMMIT_BRANCH, DRONE_TAG, DRONE_BUILD_EVENT will not be set for the container. The clone will fail when the step runs due to missing values.
</blockquote></details>

Shouldn't clone fail right away instead?

shubham149:
*[`332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java` line 726 at r9](https://reviewable.io/reviews/harness/harness-core/34679#-N7dD5kTO_jNhEG74Pui-r9-726:-N9_PsWN95o0b1oJzQ3F:b45u9is) ([raw file](https://github.com/harness/harness-core/blob/d226f796dbd1b6106bcc1b6fb26de6e34c35460d/332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java#L726)):*
<details><summary><i>Previously, ricksjames wrote…</i></summary><blockquote>

When left unset, DRONE_WORKSPACE env var will not be set for the container. When set this env variable causes the drone-git image to create this directory, and cd to it. If unset, the container will remain at the HARNESS_WORKSPACE dir "/harness".

So the clone will occur at /harness. When unset twice, both clones will try to clone to the same directory, and the second one will fail.
</blockquote></details>

Is this expected in case of 2 clones? Customer will raise bugs since the 2 clone steps are failing.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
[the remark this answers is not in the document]
shubham149: Setting variables to empty string needs to be part of PluginSettingUtils. It will work since env variables passed as part of step execute task supersedes what is passed in initialize step.

930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/CIK8InitializeTaskHandler.java
```diff
@@ -337,6 +337,11 @@ private void createEnvVariablesSecrets(CoreV1Api coreV1Api, String namespace,
...
+    Map<String, String> gitSecretData =
+        getAndUpdateGitSecretData(gitConnectorDetails, containerParamsList, k8SecretName);
+    log.info("Determined environment secrets to create for stage for pod {}", podParams.getName());
```
shubham149, line 343: Move it to L415

shubham149 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/34679)* status: Some of the checks are too old. 50 file(s) to review. 2 change requests. 16. unresolved discussions. 0 of 1 approvals obtained. Waiting on team(s): sto. 

___
*[`332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/InitializeStepUtils.java` line 40 at r4](https://reviewable.io/reviews/harness/harness-core/34679#-N7dD5kTO_jNhEG74Pum-r4-40:-N8bKj2n-j8hawhUPwmo:b-z3hxad) ([raw file](https://github.com/harness/harness-core/blob/14e073555ccc9646be98fc56c981f45bc84708c8/332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/InitializeStepUtils.java#L40)):*
<details><summary><i>Previously, ricksjames wrote…</i></summary><blockquote>

yes, 
VmInitializeStepUtils and K8InitializeStepUtils exist in this same package. These utils are used by both K8 and VM, hence I made this new class for them.
</blockquote></details>

This file is already present in develop branch. Why is it showing up as a new file here?

ricksjames:
@shubham149 Some items you mentioned do not have a reply button, so I will reference by link to comment.
IRT https://github.com/harness/harness-core/pull/34679#pullrequestreview-1073605195

InitializeStepUtils is no longer a file that is part of this PR.

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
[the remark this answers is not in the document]
ricksjames: @shubham149 
Please see K8InitializeTaskParams in my latest push. 
In the method createCIK8ContainerParams(...)
The env variables from PluginSettingUtils are put in envVars on line 304:
`envVars.putAll(containerDefinitionInfo.getEnvVars()); // Put customer input env variables`

and later in the method commonEnvVars are put in envVars on line 356 overwriting envVars from PluginSettingUtils
`envVars.putAll(commonEnvVarsCopy); // commonEnvVars needs to be put in end because they override webhook parameters`

Even if I set them to blank in PluginSettingUtils they are overwritten by the the codebase env variables that exist in commonEnvVars.

332-ci-manager/service/src/main/java/io/harness/ci/beans/steps/stepinfo/GitCloneStepInfo.java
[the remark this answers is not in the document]
ricksjames: After the bug bash we determined that this requirement should be changed. The new requirement is that the Clone Directory is optional in the UI, and that the default for clone directory is /harness/<repoName> if not supplied and /harness will not be allowed for gitCloneStep. This change has already been implemented.

332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java
ricksjames: After the bug bash we determined that this requirement should be changed. The new requirement is that the Clone Directory is optional in the UI, and that the default for clone directory is /harness/[repoName] if not supplied and /harness will not be allowed for gitCloneStep. This change has already been implemented.

ricksjames:
@shubham149 IRT https://github.com/harness/harness-core/pull/34679#issuecomment-1216219518

After the bug bash we determined that this requirement should be changed. The new requirement is that the Clone Directory is optional in the UI, and that the default for clone directory is /harness/[repoName] if not supplied and /harness will not be allowed for gitCloneStep. This change has already been implemented.

ricksjames:
@shubham149 IRT https://github.com/harness/harness-core/pull/34679#issuecomment-1216218625
Fixed.

ricksjames:
trigger compile pmd ti2 ti3 ft1 ft2

332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java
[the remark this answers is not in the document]
ricksjames: Since this logic was in K8 specific code is wasn't effective for VMs, so I did as you suggested and set these to an empty string in PluginStepUtils. The code in K8InitializeTaskParamsBuilder needed to remain for the reason mentioned above.

shubham149:
*[`332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java` line 343 at r7](https://reviewable.io/reviews/harness/harness-core/34679#-N7dD5kTO_jNhEG74Puq-r7-343:-N9kdr-M9CqYBgP-PxtC:b1k9aq9) ([raw file](https://github.com/harness/harness-core/blob/50d1f689870b594e3732a81ef078d850355744d2/332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java#L343)):*
<details><summary><i>Previously, ricksjames wrote…</i></summary><blockquote>

Since this logic was in K8 specific code is wasn't effective for VMs, so I did as you suggested and set these to an empty string in PluginStepUtils. The code in K8InitializeTaskParamsBuilder needed to remain for the reason mentioned above.
</blockquote></details>

In K8, when the step execution happens, io.harness.ci.states.AbstractStepExecutable#serialiseStep method fetches the env variables to set and it will overwrite the variables part of init step. Hence, I guess it should work for K8 as well.

shubham149:
*[`930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/CIK8InitializeTaskHandler.java` line 343 at r10](https://reviewable.io/reviews/harness/harness-core/34679#-N5v3p_B3S0W7Wq6r6rK-r10-343:-N9keYSHBWlR8e6abJGz:bcopapm) ([raw file](https://github.com/harness/harness-core/blob/48566d981d095d83690dd7409aa487ca89a2e113/930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/CIK8InitializeTaskHandler.java#L343)):*
<details><summary><i>Previously, ricksjames wrote…</i></summary><blockquote>

This section was moved on purpose to enable gitCloneStep secret data to not be overwritten by clone codebase secret data, yet allowing clone codebase secret data to remain available in all containers. 
Will not move.
</blockquote></details>

Move the log line to L415

harshjain12 (dismissed):
> Reviewed 1 of 13 files at r2, 6 of 35 files at r3, 27 of 58 files at r4, 6 of 33 files at r5, 11 of 30 files at r7, 2 of 9 files at r8, 21 of 30 files at r9, 3 of 8 files at r10, 7 of 8 files at r11, 9 of 9 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/34679)* status: 1 change requests. Waiting on team(s): sto.

332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java
```diff
@@ -662,6 +703,169 @@ private static Map<String, String> getUploadToS3StepInfoEnvVariables(UploadToS3S
     return map;
   }
 
+  private Map<String, String> getGitCloneStepInfoEnvVariables(
```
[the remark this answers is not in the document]
harshjain12: For now just carve out logic to multiple small methods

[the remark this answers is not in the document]
ricksjames: fixed

ricksjames (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/34679)* status: 6 file(s) to review. 1 change requests. 4. unresolved discussions. 0 of 1 approvals obtained. Waiting on team(s): sto. 

___
*[`332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java` line 718 at r9](https://reviewable.io/reviews/harness/harness-core/34679#-N7dD5kTO_jNhEG74Pui-r9-718:-N9mXjs7DFWmRVf4dRUE:b-7ql1td) ([raw file](https://github.com/harness/harness-core/blob/d226f796dbd1b6106bcc1b6fb26de6e34c35460d/332-ci-manager/service/src/main/java/io/harness/ci/execution/buildstate/PluginSettingUtils.java#L718)):*
<details><summary><i>Previously, shubham149 (Shubham Agrawal) wrote…</i></summary><blockquote>

Shouldn't clone fail right away instead?
</blockquote></details>

fixed
___
*[`332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/InitializeStepUtils.java` line 40 at r4](https://reviewable.io/reviews/harness/harness-core/34679#-N7dD5kTO_jNhEG74Pum-r4-40:-N9mXnGjBEDDQ7eNfOz7:b-ugdtr) ([raw file](https://github.com/harness/harness-core/blob/14e073555ccc9646be98fc56c981f45bc84708c8/332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/InitializeStepUtils.java#L40)):*
<details><summary><i>Previously, shubham149 (Shubham Agrawal) wrote…</i></summary><blockquote>

This file is already present in develop branch. Why is it showing up as a new file here?
</blockquote></details>

this file is no longer part of this PR
___
*[`332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java` line 343 at r7](https://reviewable.io/reviews/harness/harness-core/34679#-N7dD5kTO_jNhEG74Puq-r7-343:-N9mXSaWDZxQbVT6zg2j:b-z0lgwk) ([raw file](https://github.com/harness/harness-core/blob/50d1f689870b594e3732a81ef078d850355744d2/332-ci-manager/service/src/main/java/io/harness/ci/execution/integrationstage/K8InitializeTaskParamsBuilder.java#L343)):*
<details><summary><i>Previously, shubham149 (Shubham Agrawal) wrote…</i></summary><blockquote>

In K8, when the step execution happens, io.harness.ci.states.AbstractStepExecutable#serialiseStep method fetches the env variables to set and it will overwrite the variables part of init step. Hence, I guess it should work for K8 as well.
</blockquote></details>

Yes, you are correct. Fixed as suggested and working.
___
*[`930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/CIK8InitializeTaskHandler.java` line 343 at r10](https://reviewable.io/reviews/harness/harness-core/34679#-N5v3p_B3S0W7Wq6r6rK-r10-343:-N9mXhDXDBsDyy36rSSy:b-7ql1td) ([raw file](https://github.com/harness/harness-core/blob/48566d981d095d83690dd7409aa487ca89a2e113/930-delegate-tasks/src/main/java/io/harness/delegate/task/citasks/cik8handler/CIK8InitializeTaskHandler.java#L343)):*
<details><summary><i>Previously, shubham149 (Shubham Agrawal) wrote…</i></summary><blockquote>

Move the log line to L415
</blockquote></details>

fixed

dgarg-harness (dismissed):
> Reviewed 3 of 58 files at r4, 4 of 33 files at r5, 2 of 30 files at r7, 1 of 9 files at r8, 20 of 30 files at r9, 3 of 8 files at r10, 6 of 8 files at r11, 5 of 9 files at r13, 6 of 6 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/34679)* status: 1 change requests. 4. unresolved discussions. Waiting on team(s): sto.

ricksjames:
trigger compile ti1 ft1 ft2

dgarg-harness:
@shubham149 @harshjain12 - I am going ahead with the merge here since @ricksjames has addressed the review comments. He can take care of any follow up comments in a new PR.

dgarg-harness (approved):
> Reviewed 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/34679)* status: 1 change requests. 4. unresolved discussions. Waiting on team(s): sto.

dgarg-harness (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/34679#-:-N9n18r74mf5f9TP7Jds:bnfp4nl)* status: 1 change requests. 4. unresolved discussions. Waiting on team(s): sto.

PR #35157, opened 2022-07-08
[fix]: [PL-26397]: Remove Input Set validity check on Pipeline Update for Remote Pipelines

namanvermah:
Fixes:
* For remote pipelines, the input sets validation done on pipeline update should not happen.
* When creating an input set, we should store the created YAML in DB so that future input set update events have a previous yaml to compare to (already doing this for pipelines, missed out for input sets).
* For input sets list view, we should generate the error response only if we know that there is an error. To check this, the isInvalid flag can be used. This is flag can be true only for inline input sets.

This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/35157)


archit-harness (approved):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35157)* status: 5 file(s) to review. 1 of 2 approvals obtained.

mohitg0795 (approved):
:lgtm:

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35157#-:-N6SQBOm3-30KE3zu13w:bnfp4nl)* status: :shipit: complete! ✓

PR #35342, opened 2022-07-14
[fix]: [PIE-3831]: Use yaml without appended validators for schema validation

namanvermah:
For schema validations, we don't want input set validators to be appended. For example, if some timeout field in the pipeline is `<+input>.allowedValues(12h, 1d)`, and the runtime input gives a value `12h`, the value for this field in pipeline yaml to be executed will be `12h.allowedValues(12h, 1d)` for validation during execution. However, this value will give an error in schema validation. That's why we need a value that doesn't have this validator appended.

This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/35342)


archit-harness (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35342)* status: 2 file(s) to review. 1 of 2 approvals obtained.

brijeshdhakar (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35342)* status: :shipit: complete! ✓

PR #35993, opened 2022-08-03
[fix]: [CDS-40139]: Read Permissions not being Respected for App Defaults

rafaelmart1ns:
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/35993)


diptiman-adak:
Quality gate Quality gate: **_Passed_** 
 


            
 No issues Everything looks good
 [Details](https://app.embold.io/organization/gh/harness/repositories/7752dbb3099d2dba3b5d37f5931a5e4c/pull-request/35993)

rafaelmart1ns:
trigger ti4

DeepakPatankar (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35993)* status: 4 file(s) to review. 1 change requests. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/UsageRestrictionsServiceImpl.java` line 145 at r1](https://reviewable.io/reviews/harness/harness-core/35993#-N8cT1ne0qv40ngbk88i:-N8cT1ne0qv40ngbk88j:b-axn9f7) ([raw file](https://github.com/harness/harness-core/blob/9b2d1a6d54b7ad81f7427c043280dbab939b127c/400-rest/src/main/java/software/wings/service/impl/UsageRestrictionsServiceImpl.java#L145)):*
> ```Java
>     if (isNotEmpty(appIdFromRequest) && !appIdFromRequest.equals(GLOBAL_APP_ID)) {
>       if (hasNoRestrictions || isScopedToAccount) {
>         if (isNotEmpty(appEnvMapFromPermissions)) {
> ```

But the entity we are trying to use hasNoRestrictions, then why we are checking for app restrictions here



_Code quote:_
```Java
hasNoRestrictions
```

rafaelmart1ns (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35993)* status: 4 file(s) to review. 1 change requests. 1. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/UsageRestrictionsServiceImpl.java` line 145 at r1](https://reviewable.io/reviews/harness/harness-core/35993#-N8cT1ne0qv40ngbk88i:-N8cqLdW8QpDvGQFHgA5:bktafut) ([raw file](https://github.com/harness/harness-core/blob/9b2d1a6d54b7ad81f7427c043280dbab939b127c/400-rest/src/main/java/software/wings/service/impl/UsageRestrictionsServiceImpl.java#L145)):*
<details><summary><i>Previously, DeepakPatankar (Deepak Patankar) wrote…</i></summary><blockquote>

But the entity we are trying to use hasNoRestrictions, then why we are checking for app restrictions here
</blockquote></details>

Because it is only admin if it has the `Manage Application` which is checked in the previous step, so it has no restrictions but is not admin, returning false in the old step. Now I simply check if it has no restrictions and if it has access to the app, if so it returns that the user has access to that information as a view and this does not impact other cases such as writing and updating.

DeepakPatankar (changes requested):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35993)* status: 2 file(s) to review. 1 change requests. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/UsageRestrictionsServiceImpl.java` line 145 at r1](https://reviewable.io/reviews/harness/harness-core/35993#-N8cT1ne0qv40ngbk88i:-N8cvNiwCXMqOjWnFfje:bkll2cl) ([raw file](https://github.com/harness/harness-core/blob/9b2d1a6d54b7ad81f7427c043280dbab939b127c/400-rest/src/main/java/software/wings/service/impl/UsageRestrictionsServiceImpl.java#L145)):*
<details><summary><i>Previously, rafaelmart1ns (Rafael Martins) wrote…</i></summary><blockquote>

Because it is only admin if it has the `Manage Application` which is checked in the previous step, so it has no restrictions but is not admin, returning false in the old step. Now I simply check if it has no restrictions and if it has access to the app, if so it returns that the user has access to that information as a view and this does not impact other cases such as writing and updating.
</blockquote></details>

I think I am confused with the definition of hasNoRestrictions. What does this variable means ?

rafaelmart1ns (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35993)* status: 2 file(s) to review. 1 change requests. 1. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/UsageRestrictionsServiceImpl.java` line 145 at r1](https://reviewable.io/reviews/harness/harness-core/35993#-N8cT1ne0qv40ngbk88i:-N8dGY_JBPv2SjDm6ZNt:bstjjea) ([raw file](https://github.com/harness/harness-core/blob/9b2d1a6d54b7ad81f7427c043280dbab939b127c/400-rest/src/main/java/software/wings/service/impl/UsageRestrictionsServiceImpl.java#L145)):*
> sed with the definition of hasNoRestrictions. What does this varia
I think your understanding is correct, just because the previous flow was weird. If you see in the list method, the appId for this request was passed as null and always gave a wrong return, and correcting this to keep the flow closer to what I created it. But I understand that this hasNoRestrictions is the same thing, only as in this specific request we don't have a parameter it will be null and consequently null. And I believe that this flow is consistent. But it's a piece of code that needs to be validated well in PR env.

Basically creating a new validation layer.

rafaelmart1ns (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35993)* status: 7 file(s) to review. 1 change requests. 1. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/UsageRestrictionsServiceImpl.java` line 145 at r1](https://reviewable.io/reviews/harness/harness-core/35993#-N8cT1ne0qv40ngbk88i:-N8jgB8j1GSgsTmE8FCN:bxyeebp) ([raw file](https://github.com/harness/harness-core/blob/9b2d1a6d54b7ad81f7427c043280dbab939b127c/400-rest/src/main/java/software/wings/service/impl/UsageRestrictionsServiceImpl.java#L145)):*
<details><summary><i>Previously, rafaelmart1ns (Rafael Martins) wrote…</i></summary><blockquote>

> sed with the definition of hasNoRestrictions. What does this varia
I think your understanding is correct, just because the previous flow was weird. If you see in the list method, the appId for this request was passed as null and always gave a wrong return, and correcting this to keep the flow closer to what I created it. But I understand that this hasNoRestrictions is the same thing, only as in this specific request we don't have a parameter it will be null and consequently null. And I believe that this flow is consistent. But it's a piece of code that needs to be validated well in PR env.

Basically creating a new validation layer.

</blockquote></details>

Can you close this, I changed the execution logic and this execution no longer exists.

srinivaswings (dismissed):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35993)* status: 1 change requests. 1. unresolved discussions.

DeepakPatankar (approved):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35993)* status: :shipit: complete! ✓

lasdox (approved):
> Reviewed 2 of 7 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/35993)* status: :shipit: complete! ✓

PR #36045, opened 2022-08-04
feat: [CDS-39985]: graphql appid required

abhinav-harness:
## Checklist

## Comment Triggers
<details>
  
</details>

<details>
  
- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/36045)


diptiman-adak:
 


            
 #### 2 Issues 
* Code issue 2 Code issues 

 [Details](https://app.embold.io/organization/gh/harness/repositories/7752dbb3099d2dba3b5d37f5931a5e4c/pull-request/36045)

lasdox:
Suggestion:
If user sends deployedService without appId we should let it go, because deployedService has higher cardinality.
Also, I think we should also stop that autopopulating appId behind the same FeatureFlag

abhinav-harness:
@lasdox appid i think is required for rbac 
so if some api key doesnt have access to some app mongo layer is filtering it out. 
Thats why I tried to make appid a required field instead

lasdox:
@abhinav-harness 

I checked the code and MongoPersistency auto populates the appIds that the user/apikey has access.
software/wings/dl/WingsMongoPersistence.java:373
If customer doesnt send appId filter, the RBAC is safe because of this.

Then the method called `addAppIdValidation` populates all appIds from the account into the query. This method seems to be only dirtying the query. I think we should remove it.

For customers that has lots of services by appId would have performance greatly improved by sending deployedServices because of the new index anurag created.

In my opinion it doesnt make sense to ask for appId if customer send deployedServices.

What do you think?

abhinav-harness:
Yes @lasdox but only issue is deployed services can be a range as well as equals which will give problems to our indexes. 
Second thing - yes we can directly scan via deployed services (and use that index) but even in that case restricting to single app doesn't take away anything and gives us more time to analyze - WDYT ?

lasdox:
@abhinav-harness 
When i said "deployedServices" what I did mean is "services"
I think the approach is good, we can continue with that.
Its just that services should be even more performatic and sending appId and service would be redundant.
But we can start with appId for now.

What do you think about removing the `addAppIdValidation` method?

abhinav-harness:
Even if we remove appid validation auth layer will add it and queries will perform poorly. Why not restrict and mandate appId?

lasdox (approved):
> Reviewed 4 of 6 files at r1, 1 of 2 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/36045)* status: :shipit: complete! ✓

PR #37017, opened 2022-09-01
feat: [DEL-4677]: Delegate platform

mdmilic:
## Describe your changes
This is a relatively big PR, but quite a lot of files changed are just moving them to separate packages or generally package renames, so that some of the big packages can be broken down into smaller bazel targets. Most of these automatic changes were contained inside separate commits for easier reviewing (e.g. you can review commit ranges separately)
Most of the actual changes are in 260-delegate and I tried to contain them inside the packages which end with '.platform' those packages are not on the regular delegate path, so should not affect existing delegates. Anything that was common between existing delegates and the new core delegate is inside packages named '.common'
The main/core business logic of delegate is **duplicated** inside 'AbstractDelegateAgentServiceImpl.java', this is intentional to reduce the blast radius during POC. After POC older delegates should carefully migrate to 'AbstractDelegateAgentServiceImpl.java' to remove duplication. There is also DelegatePlatformService which contains non-core delegate business logic needed for Shell script only delegate. This will be removed after POC. 
The core delegate business logic will need to be modified further to add more lifecycle hooks and probably to make a better split to what is core and what is not, but current state should be good enough for POC.

Some of the changes:
* AbstractDelegateService is intentionally monolithic. This is so that it's easier to reason about in this PR, but also be able to track diffs with DelegateAgentServiceImpl because it duplicates some code. Once DelegateAgentServiceImpl is migrated to extend AbstractDelegateService, we will further properly break this down.
* Created new Kryo registrars specific for core delegate which include only classes needed for acquireTask and sendResponse API 97ecc57f9468abe40ec35b0114ec76ca194ea93c
* Created core delegate specific Kryo serializer which doesn't check class packages. This is needed because of package level targets. 3ac6cc9a4dfd593c9fa275c7bb99da8fe3020a09
* Created all the package level targets that are needed for minimal (core) delegate functionality for modules 260, 420 & 930
* To be able to create some package level targets, certain classes had to be moved to a separate package (to be able to exclude unnecessary deps). This triggered a lot of refactored files (probably 95% of the PR). Biggest culprits are package changes for AbstractDelegateValidateTask DelegateConnectionResult & DelegateValidateTask 7c248c445528deb7e2d6785932112ae32ebc16c2
* Removed Spring loader as not necessary for core (it was originally added for WinRM deps, which are not in core) 6c944f833ecd0f195787e17b2bd436ca5659820b
* Split Guice modules into common, platform and "legacy" in their own packages. Common are used in old delegates as well as core. Platform modules are used only for core delegate. Legacy modules are used for old delegates and they are not changed

## Checklist

## Comment Triggers
<details>
  
</details>

<details>
  
- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)


diptiman-adak:
 


            
 #### 85 Issues 
* Vulnerability 25 Vulnerabilities 
 * Code issue 51 Code issues 
* Anti-pattern 9 Anti-patterns 

 [Details](https://app.embold.io/organization/gh/harness/repositories/7752dbb3099d2dba3b5d37f5931a5e4c/pull-request/37017)

raghuAtWings:
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/common/DelegateTokensModule.java` line 26 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NDfmo6gD1_AXwmm9ogy:-NDfmo6gD1_AXwmm9ogz:b7wee6e) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/app/modules/common/DelegateTokensModule.java#L26)):*
> ```Java
>     bind(TokenGenerator.class)
>         .toInstance(new TokenGenerator(configuration.getAccountId(), configuration.getDelegateToken()));
>     //    bind(DelegateServiceTokenHelper.class)
> ```

remove commented lines

raghuAtWings:
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java` line 50 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfr2q6Bkm0bGsH9lVD:-NDfr2q6Bkm0bGsH9lVE:b-nlwg38) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java#L50)):*
> ```Java
>   public ScheduledExecutorService healthMonitorExecutor() {
>     return new ScheduledThreadPoolExecutor(
>         20, new ThreadFactoryBuilder().setNameFormat("healthMonitor-%d").setPriority(Thread.MAX_PRIORITY).build());
> ```

we should reduce the threads here, make minimum to be 1 and max to be 20 instead of hardcoding it to 20

raghuAtWings:
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegatePlatformModule.java` line 80 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfrIoT91dtMudrGn3t:-NDfrIoT91dtMudrGn3u:b-4bygje) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegatePlatformModule.java#L80)):*
> ```Java
>   //            final String managerHostAndPort = System.getenv("MANAGER_HOST_AND_PORT");
>   //            if (isNotBlank(managerHostAndPort)) {
>   //                log.info("Running immutable delegate, starting CCM event tailer");
> ```

remove commented code

raghuAtWings:
*[`260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformConfig.java` line 42 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfr_BF16edjNwIvK7o:-NDfr_BF16edjNwIvK7p:bytq62z) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformConfig.java#L42)):*
> ```Java
>   @JsonProperty private final String clientCertificateKeyFilePath;
>   @JsonProperty private final boolean trustAllCertificates;
>   @JsonProperty private final boolean grpcAuthorityModificationDisabled;
> ```

we should have the clean one to begin with. remove the following

1. doUpgrade
2. managerTarget
3. managerAuthority
4. queueFilePath (please see if we need it)
5. grpcServiceEnabled -> check if we need it
6. grpcServiceConnectorPort -> check if we need it
7. clientToolsDownloadDisabled
8. maxCachedArtifacts
9. grpcAuthorityModificationDisabled

raghuAtWings:
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 69 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfsM0c0shIB-X9N0PU:-NDfsM0c0shIB-X9N0PV:b-ie5gxc) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L69)):*
> ```Java
> @Singleton
> @Slf4j
> public class DelegatePlatformService extends AbstractDelegateAgentServiceImpl {
> ```

If its extending ServiceImpl then we should rename it to DelegatePlatformServiceImpl

raghuAtWings (commented):
> Reviewed 215 of 269 files at r1, 124 of 124 files at r2, 8 of 13 files at r3.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 7 file(s) to review. 0 of 1 approvals obtained.

xingchi-jin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: Some of the checks are too old. 7 file(s) to review. 1 change requests. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformApplication.java` line 51 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZ4xch2Jg9gz6RTNSe:-NEZ4xch2Jg9gz6RTNSf:b-ujoxfk) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformApplication.java#L51)):*
> ```Java
> 
> @Slf4j
> public class DelegatePlatformApplication extends Application<DelegatePlatformConfig> {
> ```

Now delegate immutable, legacy and this new platform all have an application class, where immutable and this platform are dw applications.

With guice, we can just use the same application class but bind different modules based on the Configuration, in this way, we can avoid duplicate codes and apply same fix effective to both immutable and platform delegate. 
___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 2683 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZFsws8moFhhrCVW8F:-NEZFsws8moFhhrCVW8G:bx7mhts) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L2683)):*
> ```Java
> 
>   @Override
>   public void sendTaskResponse(final String taskId, final DelegateTaskResponse taskResponse) {
> ```

this will be the other part of TaskExecutionListener
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 69 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZAuM52MmzNQHEUKLg:-NEZAuM52MmzNQHEUKLh:bse3s7i) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L69)):*
> ```Java
> @Singleton
> @Slf4j
> public class DelegatePlatformService extends AbstractDelegateAgentServiceImpl {
> ```

Agree with your code logics that these callbacks should be separated out, this is awesome. But inherit is something I would avoid.

Inherit has its problems which we can avoid using it. With 2+ layers, nobody can really define clearly what does abstract class do, what does impl class do. Eventually people tend to change codes in both places which defeats the purpose of modularization and inherit. For that, we can keep the DelegateAgentServiceImpl class code, just try to refactor it.
suggest to
1. separate codes to different modules,
       for functions processing DelegateTaskPackage, it can be a TaskExecutingModule. (In this way, with bijou mode, we just need to replace with this module)
       for functions processing TaskEvents, it can be a TaskEventProcessingModule
2. define listener interfaces for callbacks
3. attach listeners to corresponding resources using association (a member of service class). For example,   class TaskExecutingService { ...  private TaskExecutingListener listener .. } 

 I will explain in the next following comments.
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 92 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZBUXi0yVSirnCY3Lv:-NEZBUXi0yVSirnCY3Lw:b-5l9qfe) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L92)):*
> ```Java
> 
>   @Override
>   protected void executeTask(@NonNull final DelegateTaskPackage delegateTaskPackage) {
> ```

It can be a TaskExecutionModule, which handles tasks that are certain to be executed. Input is DelegateTaskPackage, output is TaskResponse, it runs tasks asynchronously. 
Correspondingly, a TaskExecutionService will be ingested to DelegatePlatformService (a.k.a DelegateAgentServiceImpl)
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 125 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZCvJf4_gEhto-49ws:-NEZCvJf4_gEhto-49wt:b-6mtnuk) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L125)):*
> ```Java
>   }
> 
>   private BooleanSupplier getPreExecutionFunction(@NotNull DelegateTaskPackage delegateTaskPackage) {
> ```

public interface ExecutingTaskListener {
   preProcessingTaskPackage(@NotNull DelegateTaskPackage delegateTaskPackage)
   postExecutingTask(TaskResponse ...)
}
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 182 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZGtTGAiXomCcwvEaa:-NEZGtTGAiXomCcwvEab:b-gvkfi0) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L182)):*
> ```Java
> 
>   @Override
>   protected void abortTask(DelegateTaskAbortEvent delegateTaskEvent) {
> ```

public TaskEventListener() {
  void abortTask(DelegateTaskAbortEvent delegateTaskEvent);
  void preProcessingTaskEvent
  void onPostExecute
}
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 238 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZJPmDArS3Txllcc7d:-NEZJPmEDhktYBDQ-NBF:b-r81i7z) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L238)):*
> ```Java
> 
>   @Override
>   protected void onDelegateStart() {
> ```

this listen belongs to the resource "delegate", it can belong here
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 243 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZK2153kdQ6Yv54HuY:-NEZK2153kdQ6Yv54HuZ:b-r81i7z) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L243)):*
> ```Java
> 
>   @Override
>   protected void onDelegateRegistered() {}
> ```

this listen belongs to the resource "delegate", it can belong here
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 246 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZIrgz1wvTEY3AuSCv:-NEZIrgz1wvTEY3AuSCw:b95s0rc) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L246)):*
> ```Java
> 
>   @Override
>   protected void onHeartbeat() {
> ```

Heartbeat should put in a DelegateHeartbeatModule, but my other PR has taken care. You can leave it this wy.
___
*[`260-delegate/src/main/java/io/harness/serializer/platform/DelegatePlatformRegistrar.java` line 53 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZ3o684sjcADMl6M7l:-NEZ3o684sjcADMl6M7m:bhvjtoj) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/serializer/platform/DelegatePlatformRegistrar.java#L53)):*
> ```Java
>     kryo.register(ScriptType.class, 5253);
> 
>     // DelegateTaskResponse
> ```

to get delegate platform to run properly, the following four types of kryo registrars are needed. Need to pick the necessary ones from the four groups 
ConnectorBeansKryoRegistrar.class
DelegateTasksBeansKryoRegister.class
DelegateBeansKryoRegistrar.class
ApiServiceBeansKryoRegister.class

xingchi-jin (commented):
I just started review with 260, I will continue the review this week. My current comments may incur some changes, most parts are design pattern related changes with mostly copy-paste (move around) code blocks. Let me know if it makes sense and I am always welcome for discussions.
When I worked on testing and end to end the bijou pov and made it work, there are some bug fixes in your code, but since I merged total 5 PRs together into 1 branch for integration, now I have to manually separate out these fixes out. I prefer maybe I just pick the significant ones out and send you in a paragraph and send to you, I will do soon. 

As we discussed, please summarize your PR in bullet points, it can make sure coder and reviewers are at the same page item by item. 
Thank you for the change ~! 

> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017#-:-NEZKafL5aMENQsKiK3H:b4kzoy4)* status: Some of the checks are too old. 7 file(s) to review. 1 change requests. 0 of 1 approvals obtained.

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: Some of the checks are too old. 7 file(s) to review. 1 change requests. 3. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformApplication.java` line 51 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZ4xch2Jg9gz6RTNSe:-NEbuB9b65F9CsxSvtfS:b30z6pe) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformApplication.java#L51)):*

Now delegate immutable, legacy and this new platform all have an application class, where immutable and this platform are dw applications.

With guice, we can just use the same application class but bind different modules based on the Configuration, in this way, we can avoid duplicate codes and apply same fix effective to both immutable and platform delegate. 
</blockquote></details>

Can you point me to some resource/example of what you are referring to? It's true, the application code is pretty much the same right now for immutable and platform with only difference on what Guice modules are installed. 
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 69 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfsM0c0shIB-X9N0PU:-NEbuuBr3Q_DNkKnVRzK:b-cesj6n) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L69)):*

If its extending ServiceImpl then we should rename it to DelegatePlatformServiceImpl
</blockquote></details>

I will remove impl from the abstract class (name it AbstractDelegateAgentService)
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 69 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZAuM52MmzNQHEUKLg:-NEbv9a-7rqmHEOtmUQd:bfqmds1) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L69)):*

Agree with your code logics that these callbacks should be separated out, this is awesome. But inherit is something I would avoid.

Inherit has its problems which we can avoid using it. With 2+ layers, nobody can really define clearly what does abstract class do, what does impl class do. Eventually people tend to change codes in both places which defeats the purpose of modularization and inherit. For that, we can keep the DelegateAgentServiceImpl class code, just try to refactor it.
suggest to
1. separate codes to different modules,
       for functions processing DelegateTaskPackage, it can be a TaskExecutingModule. (In this way, with bijou mode, we just need to replace with this module)
       for functions processing TaskEvents, it can be a TaskEventProcessingModule
2. define listener interfaces for callbacks
3. attach listeners to corresponding resources using association (a member of service class). For example,   class TaskExecutingService { ...  private TaskExecutingListener listener .. } 

 I will explain in the next following comments.
</blockquote></details>

I'm planing to further break AbstractDelegateAgentService after we migrate DelegateAgentServiceImpl to start using it. Keeping it monolithic right now makes it easier to reason about and keep track of differences between those two clasess, but once they are merged we should break them further. So I agree with you, just not as part of this PR.

xingchi-jin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: Some of the checks are too old. 7 file(s) to review. 1 change requests. 1. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformApplication.java` line 51 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZ4xch2Jg9gz6RTNSe:-NEi9TKGEcax1aYTUFy0:b8r0909) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformApplication.java#L51)):*

Can you point me to some resource/example of what you are referring to? It's true, the application code is pretty much the same right now for immutable and platform with only difference on what Guice modules are installed. 
</blockquote></details>

actually many codes in DelegateApplication.java and DelegateAgentApplication.java and this file are similar. But since DelegateAgentApplication.java and this file are dw based, i m referring to these two files. 
many functions are the same codes, you can compare the codes with https://github.com/harness/harness-core/blob/develop/260-delegate/src/main/java/io/harness/delegate/app/DelegateAgentApplication.java 

i m not against having two Applications, but only vote against dup codes. So up to you with how you want to implement this. 
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 69 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZAuM52MmzNQHEUKLg:-NEi8U5PBnOd0WpkJ3Nx:bah4cva) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L69)):*

I'm planing to further break AbstractDelegateAgentService after we migrate DelegateAgentServiceImpl to start using it. Keeping it monolithic right now makes it easier to reason about and keep track of differences between those two clasess, but once they are merged we should break them further. So I agree with you, just not as part of this PR.
</blockquote></details>

sounds good ! looking forward to the break down.

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 146 file(s) to review. 1 change requests. 9. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateCommonModule.java` line 80 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfmtGJ97TPYCtYPu_4:-NFAoUbq3GYHhtmkfqYh:b-886gsx) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateCommonModule.java#L80)):*

commented lines??
</blockquote></details>

Fixed
___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java` line 50 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfr2q6Bkm0bGsH9lVD:-NFAoW4_A56NxRqO9z2t:b67klxb) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java#L50)):*

we should reduce the threads here, make minimum to be 1 and max to be 20 instead of hardcoding it to 20
</blockquote></details>

Doesn't seem possible with ScheduledExecutors that we need to use. I can set the core pool size to 1, but max is always unbounded. Keepalive for unused threads is always 10ms. Will reduce core pool size to something like 3 instead. 
___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegatePlatformModule.java` line 80 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfrIoT91dtMudrGn3t:-NFApOr3AhVbMivX9yrm:bb7f2mf) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegatePlatformModule.java#L80)):*

remove commented code
</blockquote></details>

done
___
*[`260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformApplication.java` line 51 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZ4xch2Jg9gz6RTNSe:-NFApd2FC6CU9LRNstze:b-mal07h) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformApplication.java#L51)):*

actually many codes in DelegateApplication.java and DelegateAgentApplication.java and this file are similar. But since DelegateAgentApplication.java and this file are dw based, i m referring to these two files. 
many functions are the same codes, you can compare the codes with https://github.com/harness/harness-core/blob/develop/260-delegate/src/main/java/io/harness/delegate/app/DelegateAgentApplication.java 

i m not against having two Applications, but only vote against dup codes. So up to you with how you want to implement this. 
</blockquote></details>

I understood that, but not sure what is the best way to deal with it. The logic is the same, but they use different files, e.g. config file (& java config beans), guice modules, shutdown hooks etc.. Lets deal with this later in a separate change when we understand the differences better (these files would differ a bit more later on)
___
*[`260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformConfig.java` line 42 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfr_BF16edjNwIvK7o:-NFAqOIW38fCxghVYNmf:br8z6zd) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/platform/DelegatePlatformConfig.java#L42)):*

we should have the clean one to begin with. remove the following

1. doUpgrade
2. managerTarget
3. managerAuthority
4. queueFilePath (please see if we need it)
5. grpcServiceEnabled -> check if we need it
6. grpcServiceConnectorPort -> check if we need it
7. clientToolsDownloadDisabled
8. maxCachedArtifacts
9. grpcAuthorityModificationDisabled
</blockquote></details>

Makes sense, but this config file is not used right now at all. For backwards compatibility we still load the old io.harness.delegate.configuration.DelegateConfiguration. A lot of code depends on it, so it's hard to get rid of it. Immutable delegate actually does the same. This file needs to exist though to satisfy the Dropwizard requirements (even though we ignore it). Later on I would like to move core delegate to use this file. 
___
*[`260-delegate/src/main/java/io/harness/serializer/platform/DelegatePlatformRegistrar.java` line 53 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZ3o684sjcADMl6M7l:-NFArqB-8EywRTCT-k1r:bunvdds) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/serializer/platform/DelegatePlatformRegistrar.java#L53)):*

to get delegate platform to run properly, the following four types of kryo registrars are needed. Need to pick the necessary ones from the four groups 
ConnectorBeansKryoRegistrar.class
DelegateTasksBeansKryoRegister.class
DelegateBeansKryoRegistrar.class
ApiServiceBeansKryoRegister.class


</blockquote></details>

We should create new API endpoints for acquireTask and sendTaskResponse that don't require for delegate to deserialize TaskParams and TashExecutionResponse. With that change, core delegate would not depend on these kryo registrars. We should do that in a separate PR change
___
*[`400-rest/src/main/java/io/harness/iterator/ValidationFailedTaskMessageHelper.java` line 24 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfsyBk6qwpyDssVswd:-NFAsGGDBjdWbRtagME2:b-7crdok) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/400-rest/src/main/java/io/harness/iterator/ValidationFailedTaskMessageHelper.java#L24)):*

every class that you have moved, check whether thats registered in kryo. If yes then you should ensure that we do a sanity test with older version of delegate. I would recommend to avoid changing the package name for the classes which are registered in kryo
</blockquote></details>

I've tested this behavior and package name doesn't actually play a role in Kryo, only class name. So as long as class name is the same, it doesn't matter which package it's in and it gets deserialized properly.
___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/common/DelegateTokensModule.java` line 26 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NDfmo6gD1_AXwmm9ogy:-NFAoSUM3_mMvf6vj5YX:bb7f2mf) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/app/modules/common/DelegateTokensModule.java#L26)):*

remove commented lines
</blockquote></details>

done

raghuAtWings:
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateCommonModule.java` line 80 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfmtGJ97TPYCtYPu_4:-NFERrxaBqT5T_Xa7_PF:bwytqcm) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateCommonModule.java#L80)):*

Fixed
</blockquote></details>

this is not removed

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 334 file(s) to review. 1 change requests. 16. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateCommonModule.java` line 80 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfmtGJ97TPYCtYPu_4:-NFEbLjJ7mYnKdonOOPo:b-5di0gn) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateCommonModule.java#L80)):*

this is not removed 
</blockquote></details>

The revision you linked is R2, there are further changes, so maybe that's why you don't see it, but also I removed some of the commented lines that I thought we surely don't need, but left the ones I think we will need when adding back secrets support and CI support and added 'FixMe' comment as a reminder. This is to make it easier to identify which deps we would need to add back. 
___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 2683 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZFsws8moFhhrCVW8F:-NFEc7JE27gqTLZe7YR_:b97qyy6) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L2683)):*

this will be the other part of TaskExecutionListener
</blockquote></details>

Will be done in separate PR
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 69 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZAuM52MmzNQHEUKLg:-NFEcEmnC5BesS47qtAC:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L69)):*

sounds good ! looking forward to the break down. 
</blockquote></details>

Done.
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 92 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZBUXi0yVSirnCY3Lv:-NFEcFLn9SEK9_ej88dD:b97qyy6) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L92)):*

It can be a TaskExecutionModule, which handles tasks that are certain to be executed. Input is DelegateTaskPackage, output is TaskResponse, it runs tasks asynchronously. 
Correspondingly, a TaskExecutionService will be ingested to DelegatePlatformService (a.k.a DelegateAgentServiceImpl)

</blockquote></details>

Will be done in separate PR
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 125 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZCvJf4_gEhto-49ws:-NFEcJ8e3O79oZBLmEgF:b97qyy6) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L125)):*

public interface ExecutingTaskListener {
   preProcessingTaskPackage(@NotNull DelegateTaskPackage delegateTaskPackage)
   postExecutingTask(TaskResponse ...)
}
</blockquote></details>

Will be done in separate PR
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 182 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZGtTGAiXomCcwvEaa:-NFEcJWU9cSH95y_eEE1:b97qyy6) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L182)):*

public TaskEventListener() {
  void abortTask(DelegateTaskAbortEvent delegateTaskEvent);
  void preProcessingTaskEvent
  void onPostExecute
}
</blockquote></details>

Will be done in separate PR
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 238 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZJPmDArS3Txllcc7d:-NFEcJtC8ko-CRuMLUOk:b97qyy6) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L238)):*

this listen belongs to the resource "delegate", it can belong here
</blockquote></details>

Will be done in separate PR
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 243 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZK2153kdQ6Yv54HuY:-NFEcKr70PrZDLzMYlyr:b97qyy6) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L243)):*

this listen belongs to the resource "delegate", it can belong here
</blockquote></details>

Will be done in separate PR
___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 246 at r3](https://reviewable.io/reviews/harness/harness-core/37017#-NEZIrgz1wvTEY3AuSCv:-NFEcLqu7hJPPDmBBdYJ:b97qyy6) ([raw file](https://github.com/harness/harness-core/blob/b143b0aa1fb9b27a56550701892b4feb3ba38967/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L246)):*

Heartbeat should put in a DelegateHeartbeatModule, but my other PR has taken care. You can leave it this wy.
</blockquote></details>

Will be done in separate PR

raghuAtWings:
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java` line 50 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfr2q6Bkm0bGsH9lVD:-NFEyxvjC3v2EKt0ZTZP:br9wu56) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java#L50)):*

Doesn't seem possible with ScheduledExecutors that we need to use. I can set the core pool size to 1, but max is always unbounded. Keepalive for unused threads is always 10ms. Will reduce core pool size to something like 3 instead. 
</blockquote></details>

lets not make this change with this PR

raghuAtWings (commented):
> Reviewed 2 of 269 files at r1, 4 of 13 files at r3, 140 of 140 files at r4, 10 of 10 files at r5, 209 of 209 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 1 change requests. 13. unresolved discussions. 0 of 1 approvals obtained.

xingchi-jin (changes requested):
i have few comments, mostly looks good to me.

But i have a concern regarding how to manage this delivery.
The current approach of migration you are taking is by copying code from old DelegateModule or DelegateAgentServiceImpl to new corresponding platform java classes. 
From today till the actual delivery of delegate core will still be a long runway. During this time, all fixes go into the old code will have to manually cherry-pick to platform delegate. How to manage this and make sure we don't miss something

> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017#-:-NFH8PqS8UVzPBoW_w8a:b-wpg8l5)* status: 2 file(s) to review. 1 change requests. 13. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java` line 8 at r7](https://reviewable.io/reviews/harness/harness-core/37017#-NFGyo175W49NFpkgNLV:-NFGyo175W49NFpkgNLW:b-sskn92) ([raw file](https://github.com/harness/harness-core/blob/8bd7e09f14803824caf67bc66cc3b7865a8a5d3e/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java#L8)):*
> ```Java
>  */
> 
> package io.harness.delegate.app.modules.platform;
> ```

can we apply this refactor to all delegates?
___
*[`980-commons/src/main/java/io/harness/serializer/KryoModule.java` line 67 at r7](https://reviewable.io/reviews/harness/harness-core/37017#-NFH83m952OXTMHuopyN:-NFH83m952OXTMHuopyO:b-hucgv3) ([raw file](https://github.com/harness/harness-core/blob/8bd7e09f14803824caf67bc66cc3b7865a8a5d3e/980-commons/src/main/java/io/harness/serializer/KryoModule.java#L67)):*
> ```Java
>   @Singleton
>   public KryoSerializer getKryoSerializer(final Provider<Set<Class<? extends KryoRegistrar>>> provider) {
>     return new KryoSerializer(provider.get(), false, false);
> ```

my understanding is, delPlatform will not use this serializer, instead it will use the one in PlatformKryoModule.
The only difference is the  -  KryoSerializer(provider.get(), true, false);  VS  KryoSerializer(provider.get(), false, false);

how does that help with ignoring package path change?

mdmilic (commented):
DelegateModule (used for legacy delegate) doesn't need to be keept in sync, it's not really being copied in the platform. Platform has it's own module which installs some of the same modules like DelegateModule because they have common functionality. If any changes are done in common modules it would be picked up by both apps. 
DelegateAgentServiceImpl is the only class which code is duplicated and this is intentional. This class doesn't actually have many commits. It was 1-2 per month, so "cherry-picking" them will not be a problem. 
DelegateAgentServiceImpl will soon (not in this PR) start extending AbstractDelegateService after which there will be no more duplicated code. In the PR where we make it extend the new abstract class, I plan to go through all the changes in DelegateAgentServiceImpl since Aug 16th (date I first branched) and make sure they are in. 
We need however to merge this PR ASAP, the more we wait the harder it will be to get everything back in sync.

> Reviewed 1 of 209 files at r6.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017#-:-NFOKQxoDNxn5utbp7Iz:b-mq31ed)* status: 47 file(s) to review. 1 change requests. 14. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java` line 50 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfr2q6Bkm0bGsH9lVD:-NFOKAff7WsnAw4QyV_5:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java#L50)):*

lets not make this change with this PR
</blockquote></details>

Done.

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 47 file(s) to review. 1 change requests. 15. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java` line 8 at r7](https://reviewable.io/reviews/harness/harness-core/37017#-NFGyo175W49NFpkgNLV:-NFOMceBEJsfjb-QWkaa:bi4pt5u) ([raw file](https://github.com/harness/harness-core/blob/8bd7e09f14803824caf67bc66cc3b7865a8a5d3e/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateExecutorsModule.java#L8)):*

can we apply this refactor to all delegates?
</blockquote></details>

Definitely not as part of this PR. Aim with this PR is to impact existing delegates as little as possible to reduce the risk of breaking them (e.g. that's why I opted for code duplication for DelegateAgentServiceImpl). 
After we merge this, delegates can then continue evolving separately so we can refactor old delegates if we want to. 
Personally I feel we should not invest much time in Legacy Delegate, keep it in sustaining mode only. We should invest more time in immutable and core delegates, but if we can do something to extract common modules or code we should do it as it will make it easier to maintain everything.

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 53 file(s) to review. 1 change requests. 16. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`980-commons/src/main/java/io/harness/serializer/KryoModule.java` line 67 at r7](https://reviewable.io/reviews/harness/harness-core/37017#-NFH83m952OXTMHuopyN:-NFOXgDA0fXHy18ZrwAY:bqyeu86) ([raw file](https://github.com/harness/harness-core/blob/8bd7e09f14803824caf67bc66cc3b7865a8a5d3e/980-commons/src/main/java/io/harness/serializer/KryoModule.java#L67)):*

my understanding is, delPlatform will not use this serializer, instead it will use the one in PlatformKryoModule.
The only difference is the  -  KryoSerializer(provider.get(), true, false);  VS  KryoSerializer(provider.get(), false, false);

how does that help with ignoring package path change? 
</blockquote></details>

So that flag is enabling the check which makes sure that class is registered in the correct module. This doesn't make sense for package level targets (since one package is one module) otherwise we would need to have registrar in the same package where the class is and basically one registrar == one class.

raghuAtWings:
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 69 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfsM0c0shIB-X9N0PU:-NFQJdgp08UNyOtIjXsY:b-o45cg8) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L69)):*

I will remove impl from the abstract class (name it AbstractDelegateAgentService)
</blockquote></details>

this is not done

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 55 file(s) to review. 1 change requests. 13. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java` line 69 at r2](https://reviewable.io/reviews/harness/harness-core/37017#-NDfsM0c0shIB-X9N0PU:-NFQVpToAY64WdSfCKYW:b-mmjgbv) ([raw file](https://github.com/harness/harness-core/blob/cc72312de1157b2145f14dc19801ac0ecc715a28/260-delegate/src/main/java/io/harness/delegate/service/platform/DelegatePlatformService.java#L69)):*

this is not done 
</blockquote></details>

Missed it. Done now

xingchi-jin (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 55 file(s) to review. 1 change requests. 13. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateTaskModule.java` line 21 at r10](https://reviewable.io/reviews/harness/harness-core/37017#-NFQn4beBhgZNrG88yE5:-NFQn4beBhgZNrG88yE6:b-nogods) ([raw file](https://github.com/harness/harness-core/blob/43d3c23931c45474446ec2937df46c63b35b05e2/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateTaskModule.java#L21)):*
> ```Java
> 
> @RequiredArgsConstructor
> public class DelegateTaskModule extends AbstractModule {
> ```

I want to suggest adding comment to all our delegate module classes. For example for this one, previously in legacy delegate, it's a module mainly to provide all the tasks. With core delegate, this may be the module to provide only service classes that launches k8s jobs.
A good definition by comment helps future developers to understand the refactor much easier.

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 55 file(s) to review. 1 change requests. 14. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateTaskModule.java` line 21 at r10](https://reviewable.io/reviews/harness/harness-core/37017#-NFQn4beBhgZNrG88yE5:-NFTb7GO57646XdQYJAS:be3z38b) ([raw file](https://github.com/harness/harness-core/blob/43d3c23931c45474446ec2937df46c63b35b05e2/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/DelegateTaskModule.java#L21)):*

I want to suggest adding comment to all our delegate module classes. For example for this one, previously in legacy delegate, it's a module mainly to provide all the tasks. With core delegate, this may be the module to provide only service classes that launches k8s jobs.
A good definition by comment helps future developers to understand the refactor much easier.
</blockquote></details>

I don't have a strong opinion, but I'm usually against code comments (in Java). Having comments means your code is not good, so you should rather write clean self documenting code if there is something which is not self-explanatory. Also, having comments means you need to keep code and comments in sync which never happens, so these comments might even be misleading in a year or two. Obvious exceptions are public APIs which can become Javadoc, these comments are not meant for internal usage, but external. 
That being said, I do write code comments in very rare situations and we can write comments if we decide to for this, but let's not dwell on technicalities and cosmetic issues any more in this PR. We need to close this PR ASAP as other PRs are blocked on it, but also keeping it open is extremely time consuming (it's huge, touching a lot of files so it easily gets in conflict with 'develop' branch, but also I need to do constant merges because of new tasks being added to develop - for which I need to do local re-builds and modify imports all the time)

raghuAtWings (approved):
> Reviewed 1 of 2 files at r7, 38 of 46 files at r8, 14 of 14 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 1 change requests. 13. unresolved discussions.

xingchi-jin (approved):
> Reviewed 54 of 269 files at r1, 21 of 124 files at r2, 2 of 13 files at r3, 45 of 140 files at r4, 2 of 10 files at r5, 201 of 209 files at r6, 1 of 2 files at r7, 22 of 46 files at r8, 13 of 14 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 13. unresolved discussions.

naidu-harness (approved):
> Reviewed 54 of 269 files at r1, 21 of 124 files at r2, 2 of 13 files at r3, 45 of 140 files at r4, 2 of 10 files at r5, 201 of 209 files at r6, 1 of 2 files at r7, 22 of 46 files at r8, 13 of 14 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 13. unresolved discussions.

gn-harness (approved):
> Reviewed 54 of 269 files at r1, 29 of 124 files at r2, 3 of 13 files at r3, 94 of 140 files at r4, 2 of 10 files at r5, 202 of 209 files at r6, 1 of 2 files at r7, 38 of 46 files at r8, 14 of 14 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 13. unresolved discussions.

praveensugavanam (approved):
> Reviewed 54 of 269 files at r1, 29 of 124 files at r2, 3 of 13 files at r3, 94 of 140 files at r4, 2 of 10 files at r5, 202 of 209 files at r6, 1 of 2 files at r7, 38 of 46 files at r8, 14 of 14 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37017)* status: 13. unresolved discussions.

PR #37511, opened 2022-09-14
fix: [DEL-2223]: create a new field in delegate to store tags from yaml

arpit3107:
## Checklist

## Comment Triggers
<details>
  
</details>

<details>
  
- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)


diptiman-adak:
Quality gate Quality gate: **_Passed_** 
 


            
 No issues Everything looks good
 [Details](https://app.embold.io/organization/gh/harness/repositories/7752dbb3099d2dba3b5d37f5931a5e4c/pull-request/37511)

arpit3107:
trigger ti3,ti4

xingchi-jin:
*[`420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java` line 115 at r2](https://reviewable.io/reviews/harness/harness-core/37511#-NCIsJ084G4MW6nmR_V0:-NCIsJ084G4MW6nmR_V1:br2hfef) ([raw file](https://github.com/harness/harness-core/blob/082ee7e62eb6b0ad1d4198a90a8510939ed8ffb4/420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java#L115)):*
> ```Java
>   private List<String> tags;
>   // these are tags added from delegate.yaml
>   private List<String> tagsFromYaml;
> ```

For a Delegate entity, it doesn't matter if the tag is from UI or from yaml, can we consolidate the tags and tagsFromYaml field ? Because looks like, whenever we need to read tags, we take the union. We can store the union. The benefit of storing only one tag field is, any future visits to tags can be simpler and less possible mistakes. Let me know if it makes sense.

gaurav-jain1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 3 file(s) to review. 1 change requests. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 963 at r2](https://reviewable.io/reviews/harness/harness-core/37511#-NCJBpJFDuFd89Np0OsD:-NCJBpJFDuFd89Np0OsE:b-310f9l) ([raw file](https://github.com/harness/harness-core/blob/082ee7e62eb6b0ad1d4198a90a8510939ed8ffb4/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L963)):*
> ```Java
>       delegateTags.addAll(delegate.getTags());
>     }
>     return new ArrayList<>(delegateTags);
> ```

you caller expects Set `Set<String> selectors = new HashSet<>(getUnionOfTags(delegate));`
Why we need to create ArrayList and convert back to HashSet at callers end ?



_Code quote:_
```Java
 return new ArrayList<>(delegateTags);
```
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 2700 at r2](https://reviewable.io/reviews/harness/harness-core/37511#-NCJDoBA3oPFdToHjDkX:-NCJDoBA3oPFdToHjDkY:bwmxt65) ([raw file](https://github.com/harness/harness-core/blob/082ee7e62eb6b0ad1d4198a90a8510939ed8ffb4/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L2700)):*
> ```Java
>             .keepAlivePacket(delegateParams.isKeepAlivePacket())
>             // if delegate is ng then save tags only in delegate group.
>             .tagsFromYaml(delegateParams.isNg() ? null : delegateParams.getTags())
> ```

How do we handle the case of existing running delegates here ? Their yaml tags were already be there in tags field. Are we relying for `union` operation will take care of displaying them once ? 
And how delegate work in that case ?



_Code quote:_
```Java
.tagsFromYaml(delegateParams.isNg() ? null : delegateParams.getTags())
```
___
*[`420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java` line 115 at r2](https://reviewable.io/reviews/harness/harness-core/37511#-NCIsJ084G4MW6nmR_V0:-NCJCFeFCxGsO-l9zrPz:b-8pq6oo) ([raw file](https://github.com/harness/harness-core/blob/082ee7e62eb6b0ad1d4198a90a8510939ed8ffb4/420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java#L115)):*

For a Delegate entity, it doesn't matter if the tag is from UI or from yaml, can we consolidate the tags and tagsFromYaml field ? Because looks like, whenever we need to read tags, we take the union. We can store the union. The benefit of storing only one tag field is, any future visits to tags can be simpler and less possible mistakes. Let me know if it makes sense. 
</blockquote></details>

I think idea behind this is to get metadata from UI while performing delete operation. If we are doing delete from UI/API, tags from corresponding column should get deleted. Also when we re-apply yaml with modified tags, all of the previous yaml based tags should get overwritten.
@arpit3107 correct me if I am wrong.

xingchi-jin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 3 file(s) to review. 2 change requests. 2. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java` line 115 at r2](https://reviewable.io/reviews/harness/harness-core/37511#-NCIsJ084G4MW6nmR_V0:-NCMsPxABtu_lkrMjOmx:bxuazbc) ([raw file](https://github.com/harness/harness-core/blob/082ee7e62eb6b0ad1d4198a90a8510939ed8ffb4/420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java#L115)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

I think idea behind this is to get metadata from UI while performing delete operation. If we are doing delete from UI/API, tags from corresponding column should get deleted. Also when we re-apply yaml with modified tags, all of the previous yaml based tags should get overwritten.
@arpit3107 correct me if I am wrong.
</blockquote></details>

@gaurav-jain1  Got it, I think two problems to get clarity here.
 one is what's the expected behavior for CRUD operations on tags from UI and tags from yaml.  (prod spec)
after we are clear on the first, we then can design the second, how do we store tags in DB and in-memory. (engineering spec)

 from the way I understand, we treat tags from yaml differently from tags from UI, the current implementation shows the behavior this way
1. Update/Delete tags from UI, will not update/delete tags from yaml - eg. if a tag is deleted from UI but same tag is in yaml, it'll remain
2. Update/Delete tags from yaml, will not affect tags from UI

In general, I think customers won't care where the tags are from, precisely, they don't want the trouble of managing different sources of tags, and having confusions when seeing things remaining after deletion. 
Being said, that, maybe customers do want to manage tags from multiple sources and I may miss some context here. But I'd like leave to @prateek to have a look. 
___
*[`420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java` line 115 at r2](https://reviewable.io/reviews/harness/harness-core/37511#-NCL__hz3BzpVJiU4d4g:-NCMug4UBhcm4YPWZ5b7:b-34jy62) ([raw file](https://github.com/harness/harness-core/blob/082ee7e62eb6b0ad1d4198a90a8510939ed8ffb4/420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java#L115)):*

why are we using delegate for it? why not delegate group? also i thought we moved to use delegate group for the tags. is that not the case?
</blockquote></details>

+1 on this. Better to consolidate this

xingchi-jin:
*[`420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java` line 115 at r2](https://reviewable.io/reviews/harness/harness-core/37511#-NCIsJ084G4MW6nmR_V0:-NCN4_935rUM770Oo8V9:b-mws9hk) ([raw file](https://github.com/harness/harness-core/blob/082ee7e62eb6b0ad1d4198a90a8510939ed8ffb4/420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java#L115)):*

@gaurav-jain1  Got it, I think two problems to get clarity here.
 one is what's the expected behavior for CRUD operations on tags from UI and tags from yaml.  (prod spec)
after we are clear on the first, we then can design the second, how do we store tags in DB and in-memory. (engineering spec)

 from the way I understand, we treat tags from yaml differently from tags from UI, the current implementation shows the behavior this way
1. Update/Delete tags from UI, will not update/delete tags from yaml - eg. if a tag is deleted from UI but same tag is in yaml, it'll remain
2. Update/Delete tags from yaml, will not affect tags from UI

In general, I think customers won't care where the tags are from, precisely, they don't want the trouble of managing different sources of tags, and having confusions when seeing things remaining after deletion. 
Being said, that, maybe customers do want to manage tags from multiple sources and I may miss some context here. But I'd like leave to @prateek to have a look. 
</blockquote></details>

To clarify the previous point, what I hoping to get to is just the overall thoughts about how to manage tags from both UI and yaml, and it will be great to document it, no matter for internal use or customer use.  Depending on that, if we do need two tag fields, it's ok to have them.

xingchi-jin:
*[`420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java` line 115 at r2](https://reviewable.io/reviews/harness/harness-core/37511#-NCL__hz3BzpVJiU4d4g:-NCTFlrYA42wvQHKcGFB:bfyq0va) ([raw file](https://github.com/harness/harness-core/blob/082ee7e62eb6b0ad1d4198a90a8510939ed8ffb4/420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java#L115)):*

+1 on this. Better to consolidate this
</blockquote></details>

With the discussion today, we are clear that the expected behavior is to manage tags from UI and yaml differently. Then it makes sense to me to have two fields to store them.
However, I think two points worth address here. 
1. following Raghu's comment, handles tags from delegate group
2. create a DelegateTagsHelper to encapsulate the union, update, delete logics, instead of mixing all the codes in DelegateServiceImpl, so that if in future we want to change this behavior, it will be easier and cleaner to make the change.

xingchi-jin:
*[`420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java` line 115 at r2](https://reviewable.io/reviews/harness/harness-core/37511#-NCL__hz3BzpVJiU4d4g:-NCWPByB8lHnfg2lmEEm:b-figg61) ([raw file](https://github.com/harness/harness-core/blob/082ee7e62eb6b0ad1d4198a90a8510939ed8ffb4/420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java#L115)):*
<details><summary><i>Previously, arpit3107 (Arpit Tiwari) wrote…</i></summary><blockquote>

we are not changing the existing things for CG, so we are saving only inn delegate collection.

2. yes i will carve out delegateTagsHelper just after this pr. this ticket has been pending since long hence wanted to close on this one.
</blockquote></details>

Hi Arpit, understood that time is a metric of evaluating things, adding a helper class won't cause too much change from what you already have. if it's not blocking any customers, let's do it properly, otherwise, it creates tech-debts. let me know your thoughts, we can work it out.

400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java
```diff
@@ -952,6 +951,17 @@ private List<DelegateStatus.DelegateInner> buildInnerDelegates(String accountId,
         .collect(toList());
   }
 
+  private List<String> getUnionOfTags(Delegate delegate) {
+    Set<String> delegateTags = new HashSet<>();
+    if (isNotEmpty(delegate.getTagsFromYaml())) {
```
xingchi-jin, line 956: nit: I think you can remove the two if checks

xingchi-jin (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 5 file(s) to review. 2 change requests. 3. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1218 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYoArm6sV99HY0jeK7:-NCYoArm6sV99HY0jeK8:btbym44) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1218)):*
> ```Java
>     return updatedDelegate;
>   }
> ```

add comment to this function.
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1219 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYx0B28kCuzIAIRvCs:-NCYx0B28kCuzIAIRvCt:by6wt4c) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1219)):*
> ```Java
>   }
> 
>   public Delegate updateTagsFromUI(Delegate delegate, DelegateTags delegateTags) {
> ```

if only list of tags are needed, please define arguments as minimal viable type. also please add final and @NotNull annotations to arguments that is final and not null

here it can be
 public Delegate updateTagsFromUI(final Delegate delegate, @NotNull final List<String> tags) {
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1222 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYwnWu-xhIWAM_baIY:-NCYwnWu-xhIWAM_baIZ:b-7puna4) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1222)):*
> ```Java
>     List<String> tagsFromYaml = delegate.getTagsFromYaml();
> 
>     List<String> updatedListOfTagsFromUI = removeYamlTagsWhichAreNotInUpdatedList(tagsFromYaml, delegateTags.getTags());
> ```

Why deleting yaml tags when handling UI tags ? the logic looks complicated. 
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1230 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYtzJL2_LpYKsMZ3S6:-NCYtzJL2_LpYKsMZ3S7:bqookup) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1230)):*
> ```Java
>     if (isGroupedCgDelegate(delegate)) {
>       // update for delegateGroup here
>       log.info("Updating tags for all delegates in a group {}", delegate.getDelegateGroupName());
> ```

not need to update tags from yaml here ? 
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 2765 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYw0jt-oUbv8WoUXsr:-NCYw0jt-oUbv8WoUXss:b46801j) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L2765)):*
> ```Java
>       }
>       return delegateRegisterResponse;
>     } else {
> ```

continue with the next comment, directly move to here.

It's a bad code pattern to add boolean flags to a function argument, this means a function is overloaded with multiple tasks. It makes the code unreadable and dependencies unable to handle. On the other hand.. the "upsertDelegateOperation" is already pretty broken, so I understood why it's done this way.. but let's still try to avoid this. 
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 2827 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYvWArAzKas9DGeLOq:-NCYvWArAzKas9DGeLOr:b2zs3t2) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L2827)):*
> ```Java
> 
>   @VisibleForTesting
>   Delegate upsertDelegateOperation(Delegate existingDelegate, Delegate delegate,
> ```

since it's only used once during registration.
can you remove this  firstRequestForRegistration flag and move the function call "updateDelegateYamlTagsAfterReRegistering(registeredDelegate, delegate.getTagsFromYaml());" directly into the register function
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 2888 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYxOyT9UW5NmJCBpxx:-NCYxOyT9UW5NmJCBpxy:b7r3nck) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L2888)):*
> ```Java
>   }
> 
>   private void updateDelegateYamlTagsAfterReRegistering(Delegate registeredDelegate, List<String> yamlTags) {
> ```

use minimal viable arguments, and annotations to avoid silent null point exceptions. 
 private void updateDelegateYamlTagsAfterReRegistering(@NotNull final String accountId, @NotNull final String delegateId, @NotNull final List<String> yamlTags)

xingchi-jin (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 6 file(s) to review. 2 change requests. 3. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1220 at r7](https://reviewable.io/reviews/harness/harness-core/37511#-NCxXWVw1ApIzZmj1ZXZ:-NCxXWVw1ApIzZmj1ZX_:b-gs7beg) ([raw file](https://github.com/harness/harness-core/blob/391d0907056fd64672b55b02fb9f586f1ca625d0/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1220)):*
> ```Java
>   public Delegate updateTagsFromUI(Delegate delegate, DelegateTags delegateTags) {
>     List<String> tagsFromYaml =
>         delegate.getTagsFromYaml() != null ? delegate.getTagsFromYaml() : Collections.emptyList();
> ```

Nit: It could also be



_Code snippet:_
```Java
 Optional.ofNullable(delegate.getTagsFromYaml()).orElseGet(() -> Collections.emptyList())
```
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1222 at r7](https://reviewable.io/reviews/harness/harness-core/37511#-NCxYFxZ4woIUG3RD-L3:-NCxYFxZ4woIUG3RD-L4:bhuujgu) ([raw file](https://github.com/harness/harness-core/blob/391d0907056fd64672b55b02fb9f586f1ca625d0/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1222)):*
> ```Java
>         delegate.getTagsFromYaml() != null ? delegate.getTagsFromYaml() : Collections.emptyList();
>     List<String> tagsFromUI = delegateTags.getTags();
> ```

you can consider using Set instead of List for generating the two updatedList..  .  In case the tags grow too many, especially the ones from yaml could grow huge when customers using some script to populate the tags.
but up to you for this optimization

gaurav-jain1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 6 file(s) to review. 2 change requests. 4. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 2827 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYvWArAzKas9DGeLOq:-NCyFNqj0XJ0ZFjxn__-:b-dnln0a) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L2827)):*

since it's only used once during registration.
can you remove this  firstRequestForRegistration flag and move the function call "updateDelegateYamlTagsAfterReRegistering(registeredDelegate, delegate.getTagsFromYaml());" directly into the register function
</blockquote></details>

I thought Xingchi's comment to get rid of this flag itself. You can call this method from the caller directly. 
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1224 at r7](https://reviewable.io/reviews/harness/harness-core/37511#-NCyEUmn-4-KliYJGtfH:-NCyEUmn-4-KliYJGtfI:b-ew9u15) ([raw file](https://github.com/harness/harness-core/blob/391d0907056fd64672b55b02fb9f586f1ca625d0/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1224)):*
> ```Java
> 
>     // keep those tags which are in
>     List<String> updatedListOfTagsFromYaml =
> ```

If user deleted a tag which is present via yaml, then we will delegate that from our list. But whenever delegate re-registers, that tag will appear again. 
Is that right assumption to have ?



_Code quote:_
```Java
    List<String> updatedListOfTagsFromYaml =
        tagsFromYaml.stream().filter(tag -> tagsFromUI.contains(tag)).collect(toList());
```

xingchi-jin (dismissed):
LG after resolving existing comments. And please document on current behavior of tagsFromyaml and tagsFromUI

> Reviewed 1 of 3 files at r2, 1 of 2 files at r5, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511#-:-ND1ovIY2zSphHrX1Ye1:b1c5mdn)* status: 1 change requests. 4. unresolved discussions.

arpit3107 (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 1 change requests. 11. unresolved discussions. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 956 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NBuy2gFHlgOE43n0aEG-r5-956:-NCu8Y4y-rUba70YFz8q:b5b23vi) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L956)):*

nit: I think you can remove the two if checks 
</blockquote></details>

checks are mainly for NotNull, but added for notEmpty also
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1218 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYoArm6sV99HY0jeK7:-ND1huDxCFuCIJWxZqHm:b2nhysk) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1218)):*

add comment to this function.
</blockquote></details>

ack
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1219 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYx0B28kCuzIAIRvCs:-ND1hwoP0h3UENP1viRd:b2nhysk) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1219)):*

if only list of tags are needed, please define arguments as minimal viable type. also please add final and @NotNull annotations to arguments that is final and not null

here it can be
 public Delegate updateTagsFromUI(final Delegate delegate, @NotNull final List<String> tags) {
</blockquote></details>

ack
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1222 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYwnWu-xhIWAM_baIY:-ND1jLQ5EFnZv721mw8h:b-svriig) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1222)):*

Why deleting yaml tags when handling UI tags ? the logic looks complicated. 
</blockquote></details>

because UI always sends a list of updated tags to backend to update in db. Now backend should remove the yamlTags also.

Since tags added through yaml should have option to 
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1220 at r7](https://reviewable.io/reviews/harness/harness-core/37511#-NCxXWVw1ApIzZmj1ZXZ:-ND1iTLF-rIS6DnegE_F:b2nhysk) ([raw file](https://github.com/harness/harness-core/blob/391d0907056fd64672b55b02fb9f586f1ca625d0/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1220)):*

Nit: It could also be
</blockquote></details>

ack
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1222 at r7](https://reviewable.io/reviews/harness/harness-core/37511#-NCxYFxZ4woIUG3RD-L3:-ND1ifSs4ocURLFWnpWA:bks85s7) ([raw file](https://github.com/harness/harness-core/blob/391d0907056fd64672b55b02fb9f586f1ca625d0/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1222)):*

you can consider using Set instead of List for generating the two updatedList..  .  In case the tags grow too many, especially the ones from yaml could grow huge when customers using some script to populate the tags.
but up to you for this optimization 
</blockquote></details>

yes our initial design was bad, we should have used Set<String> instead of List<String> for storing tags in db. We can take this is follow up pr
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1224 at r7](https://reviewable.io/reviews/harness/harness-core/37511#-NCyEUmn-4-KliYJGtfH:-ND1j9Bb6Wo3HUcrvaqT:b5v0138) ([raw file](https://github.com/harness/harness-core/blob/391d0907056fd64672b55b02fb9f586f1ca625d0/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1224)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

If user deleted a tag which is present via yaml, then we will delegate that from our list. But whenever delegate re-registers, that tag will appear again. 
Is that right assumption to have ?
</blockquote></details>

Yes, yamlTags will reappear while re-registering even if you have deleted them through UI.

arpit3107 (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 2 file(s) to review. 1 change requests. 11. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java` line 115 at r2](https://reviewable.io/reviews/harness/harness-core/37511#-NCL__hz3BzpVJiU4d4g:-ND22aL0A65gljzDeDCA:b-998563) ([raw file](https://github.com/harness/harness-core/blob/082ee7e62eb6b0ad1d4198a90a8510939ed8ffb4/420-delegate-service/src/main/java/io/harness/delegate/beans/Delegate.java#L115)):*

Hi Arpit, understood that time is a metric of evaluating things, adding a helper class won't cause too much change from what you already have. if it's not blocking any customers, let's do it properly, otherwise, it creates tech-debts. let me know your thoughts, we can work it out. 
</blockquote></details>

will pick it up just after this pr so that it wont be a techdebt

gaurav-jain1 (approved):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 2 file(s) to review. 11. unresolved discussions. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 3795 at r8](https://reviewable.io/reviews/harness/harness-core/37511#-ND2DQzx8T7_a-zVVZnt:-ND2DQzx8T7_a-zVVZnu:b-9v09vq) ([raw file](https://github.com/harness/harness-core/blob/a988cfa22c51d8e5443a9c0d5b1317ba4a1ec6b9/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L3795)):*
> ```Java
>     delegate.setDelegateProfileId(existingInactiveDelegate.getDelegateProfileId());
>     delegate.setTags(existingInactiveDelegate.getTags());
>     // should update here also
> ```

Remove this comment.



_Code quote:_
```Java
 // should update here also
```
___
*[`400-rest/src/main/java/software/wings/service/intfc/DelegateService.java` line 54 at r8](https://reviewable.io/reviews/harness/harness-core/37511#-ND2DYZV0Njv_8fDgWvj:-ND2DYZV0Njv_8fDgWvk:b82ugji) ([raw file](https://github.com/harness/harness-core/blob/a988cfa22c51d8e5443a9c0d5b1317ba4a1ec6b9/400-rest/src/main/java/software/wings/service/intfc/DelegateService.java#L54)):*
> ```Java
> import javax.validation.Valid;
> import javax.ws.rs.core.MediaType;
> import lombok.NonNull;
> ```

I think this is no where used, can be removed

xingchi-jin (dismissed):
> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 11. unresolved discussions. 

___
*[`400-rest/src/main/java/software/wings/service/intfc/DelegateService.java` line 56 at r8](https://reviewable.io/reviews/harness/harness-core/37511#-ND4Dhte3g1osgey4edV:-ND4Dhte3g1osgey4edW:b-n9k7jq) ([raw file](https://github.com/harness/harness-core/blob/a988cfa22c51d8e5443a9c0d5b1317ba4a1ec6b9/400-rest/src/main/java/software/wings/service/intfc/DelegateService.java#L56)):*
> ```Java
> import lombok.NonNull;
> import org.glassfish.jersey.media.multipart.FormDataContentDisposition;
> import org.jetbrains.annotations.NotNull;
> ```

normally prefer to use libraries from JEE than thirdparty libs



_Code snippet:_
```Java
 import javax.validation.constraints.NotNull;
```

xingchi-jin (approved):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 11. unresolved discussions.

arpit3107 (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 16. unresolved discussions. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 1230 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYtzJL2_LpYKsMZ3S6:-ND4T9Io6DUHINCAERH8:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L1230)):*

not need to update tags from yaml here ? 
</blockquote></details>

Done.
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 2765 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYw0jt-oUbv8WoUXsr:-ND4T9yv7z-QQ6NYbhbX:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L2765)):*

continue with the next comment, directly move to here.

It's a bad code pattern to add boolean flags to a function argument, this means a function is overloaded with multiple tasks. It makes the code unreadable and dependencies unable to handle. On the other hand.. the "upsertDelegateOperation" is already pretty broken, so I understood why it's done this way.. but let's still try to avoid this. 

</blockquote></details>

Done.
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 2827 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYvWArAzKas9DGeLOq:-ND4TAeR-H759xxr-zGn:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L2827)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

I thought Xingchi's comment to get rid of this flag itself. You can call this method from the caller directly. 
</blockquote></details>

Done.
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 2888 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYxOyT9UW5NmJCBpxx:-ND4TBSW32sRKkSY2swV:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L2888)):*

use minimal viable arguments, and annotations to avoid silent null point exceptions. 
 private void updateDelegateYamlTagsAfterReRegistering(@NotNull final String accountId, @NotNull final String delegateId, @NotNull final List<String> yamlTags)
</blockquote></details>

Done.
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 4569 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYyvSICNVqYVOG8KiY:-ND4TD1T6zIRHWHIdVim:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L4569)):*



if this complicated logic has to be here, you can divide the function into two functions. 

private List<String> getYamlTagsInUITagsList(@NotNull final List<String> yamlTags, @NotNull final List<String> uiTags) {
  return yamlTags.stream().filter(tag -> uiTags.contains(tag)).toList()
}

private List<String> getUITagsNotInYamlTagsList(@NotNull final List<String> uiTags, @NotNull final List<String> yamlTags) {
  return uiTags.stream().filter(tag -> !yamlTags.contains(tag)).toList()
}

in the caller, do two function calls instead of one. There is no performance cost with much more readable code. 
</blockquote></details>

Done.
___
*[`400-rest/src/main/java/software/wings/service/intfc/DelegateService.java` line 56 at r8](https://reviewable.io/reviews/harness/harness-core/37511#-ND4Dhte3g1osgey4edV:-ND4TE8o6vrcDPCQaSN6:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/a988cfa22c51d8e5443a9c0d5b1317ba4a1ec6b9/400-rest/src/main/java/software/wings/service/intfc/DelegateService.java#L56)):*

normally prefer to use libraries from JEE than thirdparty libs
</blockquote></details>

Done.

xingchi-jin (approved):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 11. unresolved discussions.

xingchi-jin (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 9. unresolved discussions. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 2888 at r5](https://reviewable.io/reviews/harness/harness-core/37511#-NCYxOyT9UW5NmJCBpxx:-ND4_CctCa7zgIC8woTM:b-fwnc7k) ([raw file](https://github.com/harness/harness-core/blob/1e5bfaf4a3347c994d4647663e05410b12eeb9ca/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L2888)):*
<details><summary><i>Previously, arpit3107 (Arpit Tiwari) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

the comment here is about to change the argument to
@NotBlank final String accountId, @NotBlank final String delegateId, @Nullable final List yamlTags)

it's not fixed.

xingchi-jin (approved):
:lgtm:
Thanks for the effort 

> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511#-:-ND6r6jPEu46jqwXS90b:b-25t0mo)* status: 5. unresolved discussions.

xingchi-jin (approved):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/37511)* status: 4. unresolved discussions.

PR #39191, opened 2022-10-25
feat: [PL-29164]: Support for dynamic configuration of iterators

Raghav-Murali:
## Describe your changes

This consists of changes based on the design approach described in this page - 
https://harness.atlassian.net/wiki/spaces/PLATFORM/pages/21228061570/Manager+Iterator+Service+and+dynamic+configuration+of+iterators

The brief overview of the changes -
1. Iterators will be started up based on the `iterator_config.json` file which will be present in the Bazel sandbox in local test setup and provided through Kubernetes volume mounted ConfigMap in K8s env. For the default `iterator_config.json` file please refer - https://github.com/harness/harness-core/pull/39325
2. A new `IteratorConfigMonitor` thread will monitor for directory modifications using the WatchService in the volume where the iterator config is mounted.
3. Once a change has been detected the thread will read the JSON file for the new iterator configurations.
4. If there is a change in the iterator config then the thread will safely terminate the current executor service of the iterator and restart the iterator with a new executor service based on the new configuration.
5. This will allow us to slowly port the iterators from the `manager` service to the new `manager-iterator` service which will be running only the iterators. Initially 5 low risk iterators will be made to run in `manager-iterator` service and they won't be running the `manager` service. Please refer - https://github.com/wings-software/ng-pr-manifests/pull/217 for the 5 iterators that will be disabled in the `manager` and enabled in `manager-iterator`. 
6. This will also allow to fine tune the iterators (thread-pool size, target-interval, thread-pool interval) dynamically without needing to bring down the pod or restart it with a new image. 

The new files pertaining to iterator configurations are added in `960-persistence/src/main/java/io/harness/iterator`. Some of the key new classes are - 

1. `IteratorBaseHandler` - Provides 3 default methods `registerIterators`, `createAndStartIterator` and `stopExecutor`. 
 - `registerIterators` to be overridden by all the iterator handlers to register their instance with the `IteratorConfigHandler`.
 - `createAndStartIterator` to be overridden by all the iterator handlers to create and start the iterator with iterator specific settings apart from the configuration provided by `iterator_config.json`
 - `stopExecutor` the default implementation provided by this base class suffices for most iterators, can be overridden if needed for any iterator specific clean-ups.
2. `IteratorPumpModeHandler` -  Will extend the `IteratorBaseHandler` methods for supporting Pump mode.
3. `IteratorLoopModeHandler` -  Will extend the `IteratorBaseHandler` methods for supporting Loop mode.
4. `IteratorExecutionHandler` - Manages the mapping of the iterator state and configuration. Starts up all the iterators and applies new configurations. It will use the `createAndStartIterator` and `stopExecutor` methods provided above to start and stop the iterators. 
5. `IteratorConfigWatcher` - Monitors for any changes in the `iterator_config.json` and applies those changes to the iterators by passing It to `IteratorExecutionHandler`.

The changes in all the 36 iterators are only pertaining to extending the `IteratorBaseHandler` and overriding the `createAndStartIterator` and `registerIterators` methods so that `IteratorConfigMonitorTask` has a clean way to manage all these iterators. There are no changes to the existing configuration of the iterators and whatever was configured currently is what is incorporated in `iterator_config.json`. 

## Comment Triggers
<details>
  
</details>

<details>
  
- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)


Raghav-Murali:
trigger feature-build

Raghav-Murali:
trigger feature-build

Raghav-Murali:
trigger feature-build

Raghav-Murali:
trigger feature-build

Raghav-Murali:
trigger feature-build

Raghav-Murali:
trigger feature-build

sahilhindwani-harness (approved):
> Reviewed 1 of 56 files at r1.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 56 file(s) to review. 1 of 2 approvals obtained.

archit-harness (approved):
> Reviewed 1 of 56 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 55 file(s) to review.

Raghav-Murali:
trigger feature-build

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java` line 33 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkBOa53HswIFOtujFZ:-NFkBOa53HswIFOtujF_:b71bhua) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java#L33)):*
> ```Java
>   @Value
>   @Builder
>   public static class DynamicIteratorConfigOption {
> ```

why is this an inner class? Also why is this part of 960-persistence?

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java` line 224 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkBue44lzsNsa1pRZp:-NFkBue44lzsNsa1pRZq:b-pca33r) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java#L224)):*
> ```Java
>             // The executor service termination is not yet complete. So add this iterator again to the back of the Q.
>             // By the time this task is retrieved again the executor shutdown should've gone through.
>             jobQ.add(iteratorTaskConfig);
> ```

it can create a back to back processing and can consume CPU. Shouldn't we do some wait before this can be processed?

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java` line 250 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkCB5L4SwgoVQP2yPn:-NFkCB5L4SwgoVQP2yPo:btlazq2) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java#L250)):*
> ```Java
>       configureWatcher();
> 
>       for (;;) {
> ```

why not while(true)

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 28 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkT-xj57B-yQpZ7644:-NFkT-xkChUuv41bYM86:bosaknp) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L28)):*
> ```Java
>     try {
>       // Wait for 800ms to allow all the current executing tasks to terminate.
>       if (!executor.awaitTermination(800, TimeUnit.MILLISECONDS)) {
> ```

why 800ms, shouldn't it be little more?

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 35 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkT6B664SeCvPcv7Kp:-NFkT6B664SeCvPcv7Kq:b-o0vv7i) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L35)):*
> ```Java
>     }
> 
>     if (executorService == null) {
> ```

can this be true? this should be an error isn't it? otherwise it will silently not do anything and we wouldn;t know.

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 20 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkTJWt08_EcsfSkaC9:-NFkTJWt08_EcsfSkaCA:b-o0vv7i) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L20)):*
> ```Java
>   @Override
>   protected void stopExecutor() {
>     if (executor == null) {
> ```

can this be true? this should be an error isn't it? otherwise it will silently not do anything and we wouldn;t know.

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 52 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkTNQf1dXLRbWmK4LM:-NFkTNQf1dXLRbWmK4LN:bgsp7xu) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L52)):*
> ```Java
> 
>   @Override
>   protected void registerIterators(IteratorConfigHandler iteratorConfigHandler) {
> ```

we should force this, we should make this class an abstract and mark these methods abstract

raghuAtWings (commented):
> Reviewed 50 of 56 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 55 file(s) to review. 0 of 2 approvals obtained.

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 55 file(s) to review. 7. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java` line 33 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkBOa53HswIFOtujFZ:-NFkwBPABZ12OT25MOH9:bty0zx3) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java#L33)):*

why is this an inner class? Also why is this part of 960-persistence? 
</blockquote></details>

The reason to have this as an inner class is because this is a helper class that is used only within the `IteratorConfigMonitorTask` class. So this provides a way of logical grouping and easy maintainability. 

Initially I had all these new classes as part of 400-rest where almost all the iterator handlers were defined. However, the `timeout` iterator which is part of 959-timeout-engine didn't have a dependency on 400-rest and adding a dependency was not working as it was resulting in many cycles. Thus, I moved it to 960-persistence where all the Mongo iterators related classes were defined and since these classes are also used to provide handling for Mongo iterators it made sense to define them here. Please let me know if you feel these classes should be in a different directory. 
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java` line 224 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkBue44lzsNsa1pRZp:-NFkyVFiEVOFySZoeTJt:bukadxv) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java#L224)):*

it can create a back to back processing and can consume CPU. Shouldn't we do some wait before this can be processed?
</blockquote></details>

The `stopExecutor()` method above has a 800ms wait time and each task in the Q will have to go through that wait time. Need to identify if 800ms is good enough of a wait. In local testing and PR env it seemed to work fine but have to collect some data in QA to determine if 800ms is right or need to increase it.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java` line 250 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkCB5L4SwgoVQP2yPn:-NFl0KeG7ChjNe4I0k7c:bun09dq) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java#L250)):*

why not while(true)
</blockquote></details>

I'm used to using for(;;) for infinite loops from my C++ background. Since `while(true)` throws a constant expression warning in C++. So I just went with for(;;), I can replace it with while(true).
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 20 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkTJWt08_EcsfSkaC9:-NFl2Ih6ACwhbNFQOhAe:b-e3iicx) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L20)):*

can this be true? this should be an error isn't it? otherwise it will silently not do anything and we wouldn;t know.
</blockquote></details>

This can be true if the child class that is extending this class doesn't set it. I have updated all the iterator handlers to set this field in the createAndStartIterator() so it won't be true for them. I can have it throw a NullPointerException in case any new iterators are being added and it doesn't set this.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 28 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkT-xj57B-yQpZ7644:-NFl4oZp-2_aYC809mJ3:b-dkmao0) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L28)):*

why 800ms, shouldn't it be little more?
</blockquote></details>

Yes, I didn't have sufficient data on what would be the right time to set for this. While testing in local env 800ms seemed to give good enough time to shutdown the executor services but in local env the Q doesn't have much tasks waiting to be picked. So I thought I will fine tune this value once I test it in QA env. I can have it at 2 secs (~2x 800ms) for now, let me know if you'd prefer to have this set to a even higher value.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 35 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkT6B664SeCvPcv7Kp:-NFl4eT081Iek3sBpAZC:b-e3iicx) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L35)):*

can this be true? this should be an error isn't it? otherwise it will silently not do anything and we wouldn;t know.
</blockquote></details>

This can be true if the child class that is extending this class doesn't set it. I have updated all the iterator handlers to set this field in the createAndStartIterator() so it won't be true for them. I can have it throw a NullPointerException in case any new iterators are being added and it doesn't set this.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 52 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkTNQf1dXLRbWmK4LM:-NFl5f1p-l9ZB1n7T3xG:bqvw42a) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L52)):*

we should force this, we should make this class an abstract and mark these methods abstract
</blockquote></details>

Done, will update.

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 55 file(s) to review. 7. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 20 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkTJWt08_EcsfSkaC9:-NFp1JoeBHaUyDq0MFTK:b-7if75l) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L20)):*

This can be true if the child class that is extending this class doesn't set it. I have updated all the iterator handlers to set this field in the createAndStartIterator() so it won't be true for them. I can have it throw a NullPointerException in case any new iterators are being added and it doesn't set this.
</blockquote></details>

This can be true in case when the iterator was disabled during the startup but it needs to be enabled later. This use case is applicable to the `manager-iterator` service when we slowly start to enable iterators. Thus, since it was disabled during start-up the executorService would not have been initialized. So we will shutdown the executor if its initialized and if its not yet created then we will not do anything.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 35 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkT6B664SeCvPcv7Kp:-NFp0_bUC7y1tnaGOFHk:b6vora4) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L35)):*

This can be true if the child class that is extending this class doesn't set it. I have updated all the iterator handlers to set this field in the createAndStartIterator() so it won't be true for them. I can have it throw a NullPointerException in case any new iterators are being added and it doesn't set this.
</blockquote></details>

This can be true in case when the iterator was disabled during the startup but it needs to be enabled later. This use case is applicable to the `manager-iterator` service when we slowly start to enable iterators. Thus, since it was disabled during start-up the executorService would not have been initialized. So we will shutdown the executor service if its initialized and if its not yet created then we will not do anything.

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 20 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkTJWt08_EcsfSkaC9:-NG2vvvx0-MeXDKkRufw:bxwqp3q) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L20)):*

This can be true in case when the iterator was disabled during the startup but it needs to be enabled later. This use case is applicable to the `manager-iterator` service when we slowly start to enable iterators. Thus, since it was disabled during start-up the executorService would not have been initialized. So we will shutdown the executor if its initialized and if its not yet created then we will not do anything.
</blockquote></details>

we should add an error log in that case

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 35 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkT6B664SeCvPcv7Kp:-NG2w0DP520rTJH3qXV1:byancvl) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L35)):*

This can be true in case when the iterator was disabled during the startup but it needs to be enabled later. This use case is applicable to the `manager-iterator` service when we slowly start to enable iterators. Thus, since it was disabled during start-up the executorService would not have been initialized. So we will shutdown the executor service if its initialized and if its not yet created then we will not do anything.
</blockquote></details>

add an error log in that case

raghuAtWings (commented):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 55 file(s) to review. 0 of 2 approvals obtained.

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: Some of the checks are too old. 55 file(s) to review. 2. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 20 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkTJWt08_EcsfSkaC9:-NG3iUTq4ej3dRPnBBho:bbqpgtt) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L20)):*

we should add an error log in that case
</blockquote></details>

Done, will add it.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 35 at r2](https://reviewable.io/reviews/harness/harness-core/39191#-NFkT6B664SeCvPcv7Kp:-NG3iWni4AuydHbXUbp5:bbqpgtt) ([raw file](https://github.com/harness/harness-core/blob/2ec6ec5dd9531daf7178f384c8dbfd06f09f156b/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L35)):*

add an error log in that case
</blockquote></details>

Done, will add it.

puthrayaharness (approved):

gn-harness (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: Some of the checks are too old. 56 file(s) to review. 1 change requests. 2. unresolved discussions. 1 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/yaml/gitSync/GitSyncPollingIterator.java` line 37 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGMq6nf1HPvT2RlWwCW:-NGMq6nf1HPvT2RlWwCX:b-p1eo0t) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/400-rest/src/main/java/software/wings/yaml/gitSync/GitSyncPollingIterator.java#L37)):*
> ```Java
> @OwnedBy(DX)
> public class GitSyncPollingIterator
>     extends IteratorBaseHandler implements MongoPersistenceIterator.Handler<YamlGitConfig> {
> ```

Why some are extending IteratorCustomExecutorBaseHandler and others IteratorBaseHandler



_Code quote:_
```Java
IteratorBaseHandler
```
___
*[`959-timeout-engine/src/main/java/io/harness/timeout/TimeoutEngine.java` line 149 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGMp-pE1NpvuurN0hTN:-NGMp-pE1NpvuurN0hTO:b-vkrhui) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/959-timeout-engine/src/main/java/io/harness/timeout/TimeoutEngine.java#L149)):*
> ```Java
>   @Override
>   protected void stopExecutor() {
>     super.stopExecutor();
> ```

Why is this overriden?



_Code quote:_
```Java
  protected void stopExecutor() {
    super.stopExecutor();
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 13 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGLvHAT6ILB_0TuHEkH:-NGLvHAT6ILB_0TuHEkI:b-mt3hu8) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L13)):*
> ```Java
> @Slf4j
> public abstract class IteratorBaseHandler<T extends PersistentIterable, F extends FilterExpander> {
>   public String iteratorName;
> ```

nit: create protected setIteratorName function and mark this private / or just mark it to be protected



_Code quote:_
```Java
public String iteratorName
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 17 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGLwO6y75ZQI4878ipQ:-NGLwO6y75ZQI4878ipR:b-jad602) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L17)):*
> ```Java
> 
>   /**
>    * This method returns true if executor service is terminated and vice vers.
> ```

versa*
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 40 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGLwVZk52qFW6qKSBBM:-NGLwVZk52qFW6qKSBBN:b3zpdf) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L40)):*
> ```Java
>       // The iterator is not yet initialized, so this iterator is
>       // starting for the very first time, thus return.
>       log.error("Executor not yet initialized");
> ```

nit: please add the iterator name



_Code quote:_
```Java
Executor not yet initialized");
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigHandler.java` line 9 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGLxWo79UosKDSnQTa3:-NGLxWo79UosKDSnQTa4:b-dhq1px) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorConfigHandler.java#L9)):*
> ```Java
>   private HashMap<String, IteratorBaseHandler> iteratorHandlerMap;
>   private IteratorConfigExecutorService executorService;
>   private String iteratorConfigPath;
> ```

mark these final



_Code quote:_
```Java
  private HashMap<String, IteratorBaseHandler> iteratorHandlerMap;
  private IteratorConfigExecutorService executorService;
  private String iteratorConfigPath;
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java` line 22 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGN0fRsChD5DnOtsRBs:-NGN0fRsChD5DnOtsRBt:b250vud) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java#L22)):*
> ```Java
> 
> @Slf4j
> public class IteratorConfigMonitorTask implements Runnable {
> ```

unit tests?



_Code quote:_
```Java
public class IteratorConfigMonitorTask 
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java` line 242 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGN0Y4B0KdQIaHyniUf:-NGN0Y4B0KdQIaHyniUg:bwvg1ha) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java#L242)):*
> ```Java
>    */
>   @Override
>   public void run() {
> ```

nit: public methods should come first, followed by protected and then private



_Code quote:_
```Java
blic void run() {
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 11 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGMpG3NC1GJICSX_wrL:-NGMpG3NC1GJICSX_wrM:b7wodzb) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L11)):*
> ```Java
> @Slf4j
> public abstract class IteratorCustomExecutorBaseHandler extends IteratorBaseHandler {
>   public ExecutorService executor;
> ```

nit: can these be protected?



_Code quote:_
```Java
public
```

inderschera (changes requested):
> Reviewed 46 of 56 files at r1, 1 of 6 files at r2, 1 of 3 files at r3, 8 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: Some of the checks are too old. 8 file(s) to review. 2 change requests. 2. unresolved discussions. 1 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java` line 87 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGN2-UABgf840z8pryy:-NGN2-UABgf840z8pryz:bqozvj9) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java#L87)):*
> ```Java
>                                       .in(Arrays.asList(true, null))
>                                       .and(query.criteria(ArtifactStreamKeys.collectionStatus)
>                                                .notEqual(ArtifactStreamCollectionStatus.STOPPED)))
> ```

we need to remove collectionStatus filter here. Please rebase once with develop. It's removed from develop because of performance issues.
___
*[`400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java` line 103 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGN2DCF5ZKiSNGuUeKn:-NGN2DCF5ZKiSNGuUeKo:bqy7qja) ([raw file](https://github.com/harness/harness-core/blob/8c1cef4602ac43c7b718af205f15f695152efcd5/400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java#L103)):*
> ```Java
>       artifactCollectionExecutor.scheduleAtFixedRate(
>           () -> iterator.process(), random.nextInt(10), 10, TimeUnit.SECONDS);
>     }
> ```

We added a jitter before iterator starts. This was to avoid iterator to start at the same time in all the pods. This has also been removed. Is this expected?
___
*[`400-rest/src/main/java/io/harness/workers/background/iterator/ArtifactCleanupHandler.java` line 85 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGN2dO07K1jYyHVwcr6:-NGN2dO07K1jYyHVwcr7:bwyo3mb) ([raw file](https://github.com/harness/harness-core/blob/8c1cef4602ac43c7b718af205f15f695152efcd5/400-rest/src/main/java/io/harness/workers/background/iterator/ArtifactCleanupHandler.java#L85)):*
> ```Java
>     if (iterator != null) {
>       artifactCollectionExecutor.scheduleAtFixedRate(() -> iterator.process(), 0, 5, TimeUnit.MINUTES);
>     }
> ```

same here. We removed the jitter. Is this expected?

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: Some of the checks are too old. 8 file(s) to review. 2 change requests. 9. unresolved discussions. 1 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java` line 87 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGN2-UABgf840z8pryy:-NGN_mWf2XKFe_jIsUuR:bfkpe8r) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java#L87)):*
<details><summary><i>Previously, inderschera (Inderpreet Chera) wrote…</i></summary><blockquote>

we need to remove collectionStatus filter here. Please rebase once with develop. It's removed from develop because of performance issues.
</blockquote></details>

Got it, will rebase and remove it. Thanks.
___
*[`400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java` line 103 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGN2DCF5ZKiSNGuUeKn:-NGNapTR4AKsPYxhZTc2:b-8ywili) ([raw file](https://github.com/harness/harness-core/blob/8c1cef4602ac43c7b718af205f15f695152efcd5/400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java#L103)):*
<details><summary><i>Previously, inderschera (Inderpreet Chera) wrote…</i></summary><blockquote>

We added a jitter before iterator starts. This was to avoid iterator to start at the same time in all the pods. This has also been removed. Is this expected?
</blockquote></details>

So if you go to createAndStartIterator() function above we create a Pump iterator using createPumpIteratorWithDedicatedThreadPool() and this will add a Jitter of the given targetInterval seconds. It is same as the recent fix that you made to have a jitter between 0 to 10 secs. 
___
*[`400-rest/src/main/java/io/harness/workers/background/iterator/ArtifactCleanupHandler.java` line 85 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGN2dO07K1jYyHVwcr6:-NGNbp19CeXn_7MDLIKs:bwk1zi4) ([raw file](https://github.com/harness/harness-core/blob/8c1cef4602ac43c7b718af205f15f695152efcd5/400-rest/src/main/java/io/harness/workers/background/iterator/ArtifactCleanupHandler.java#L85)):*
<details><summary><i>Previously, inderschera (Inderpreet Chera) wrote…</i></summary><blockquote>

same here. We removed the jitter. Is this expected?
</blockquote></details>

Same as above. createPumpIteratorWithDedicatedThreadPool() will add the jitter of given targetInterval.
___
*[`400-rest/src/main/java/software/wings/yaml/gitSync/GitSyncPollingIterator.java` line 37 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGMq6nf1HPvT2RlWwCW:-NGNcRnk5CF8BGWnKsh5:b-aechzx) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/400-rest/src/main/java/software/wings/yaml/gitSync/GitSyncPollingIterator.java#L37)):*

Why some are extending IteratorCustomExecutorBaseHandler and others IteratorBaseHandler
</blockquote></details>

So the Pump mode uses only one executor whereas Loop mode uses two executors. The iterators with one executor will extend IteratorBaseHandler and the iterators with two executors will extend IteratorCustomExecutorBaseHandler. Both these classes will shutdown the respective executors.
___
*[`959-timeout-engine/src/main/java/io/harness/timeout/TimeoutEngine.java` line 149 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGMp-pE1NpvuurN0hTN:-NGNeDhuD2MZ7AnP07uw:bmdua85) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/959-timeout-engine/src/main/java/io/harness/timeout/TimeoutEngine.java#L149)):*

Why is this overriden?
</blockquote></details>

This is a Loop mode iterator with 2 executors. The IteratorCustomExecutorBaseHandler expects one of the executor to be of type ScheduledThreadPoolExecutor, however this one uses SingleThreadPoolExecutor for one of its executors. Thus, overriding stopExecutor() method to shutdown the SingleThreadPoolExecutor.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 13 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGLvHAT6ILB_0TuHEkH:-NGNg_1W3Y4tVF3-E71q:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L13)):*

nit: create protected setIteratorName function and mark this private / or just mark it to be protected
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 17 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGLwO6y75ZQI4878ipQ:-NGNg_s948aamO_AF6Tm:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L17)):*

versa*
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 40 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGLwVZk52qFW6qKSBBM:-NGNiGKz2EskQcGPvPyE:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L40)):*

nit: please add the iterator name
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigHandler.java` line 9 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGLxWo79UosKDSnQTa3:-NGNjf7pAMWVRN7ONzg6:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorConfigHandler.java#L9)):*

mark these final
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java` line 22 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGN0fRsChD5DnOtsRBs:-NGNpAwL1v5rF3NvnyjE:bluoejb) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java#L22)):*

unit tests?
</blockquote></details>

I went to local and PR testing directly. Will add a few unit tests for this class.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java` line 242 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGN0Y4B0KdQIaHyniUf:-NGNp1y2-J4VyLCVNbQY:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorConfigMonitorTask.java#L242)):*

nit: public methods should come first, followed by protected and then private
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java` line 11 at r4](https://reviewable.io/reviews/harness/harness-core/39191#-NGMpG3NC1GJICSX_wrL:-NGNq8M669iND3dqg5g1:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/9aecb37c645f7a98275e47e3b3897a818006dc35/960-persistence/src/main/java/io/harness/iterator/IteratorCustomExecutorBaseHandler.java#L11)):*

nit: can these be protected?
</blockquote></details>

Done

Raghav-Murali:
trigger feature-build

960-persistence/src/main/java/io/harness/iterator/IteratorConfigHandler.java
```diff
@@ -0,0 +1,47 @@
+package io.harness.iterator;
+
+import java.util.HashMap;
+import java.util.concurrent.TimeUnit;
+
+public class IteratorConfigHandler {
```
gn-harness, line 6: At a high level, I think there should be these entities with the following behavior:

1) IteratorExecutionHandler
 - registering iterators
 - maintain state
 - start/stop
2) IteratorConfigWatcher
- Once started, looks for change in iterator configuration
- If a change is found, tell iterator execution handler about the new config
3) A pair of classes for Pump and Loop iterator configurations / these can extend from the base handler.

Raghav-Murali: Done, I will re-factor the classes and raise a newer version tomorrow.

Raghav-Murali:
trigger feature-build

archit-harness (dismissed):
> Reviewed 1 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 46 file(s) to review. 2 change requests. 10. unresolved discussions. 1 of 2 approvals obtained.

inderschera (approved):
> Reviewed 42 of 48 files at r7, 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 46 file(s) to review. 1 change requests. 7. unresolved discussions.

gn-harness (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 46 file(s) to review. 1 change requests. 7. unresolved discussions. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 31 at r8](https://reviewable.io/reviews/harness/harness-core/39191#-NGbULOa-10OwrUMToDr:-NGbULOb7M2LXIMLl96B:by5na67) ([raw file](https://github.com/harness/harness-core/blob/edae3242cbbaeeb1fb2e2b2017f90260e8501bad/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L31)):*
> ```Java
>    * @param iteratorConfigWatcher the handler for iterator configurations
>    */
>   protected abstract void registerIterators(IteratorConfigWatcher iteratorConfigWatcher);
> ```

nit: registerIterator*



_Code quote:_
```Java
registerIterators(
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigWatcher.java` line 21 at r8](https://reviewable.io/reviews/harness/harness-core/39191#-NGbU_ci4brtOaixoXRA:-NGbU_ci4brtOaixoXRB:bt0uq4e) ([raw file](https://github.com/harness/harness-core/blob/edae3242cbbaeeb1fb2e2b2017f90260e8501bad/960-persistence/src/main/java/io/harness/iterator/IteratorConfigWatcher.java#L21)):*
> ```Java
>   private final String iteratorConfigFile;
> 
>   public IteratorConfigWatcher(String iteratorConfigPath, String iteratorConfigFile) {
> ```

Add some comments on what these mean?



_Code quote:_
```Java
String iteratorConfigPath, String iteratorConfigFile
```
___
*[`960-persistence/src/test/java/io/harness/iterator/IteratorExecutionHandlerTest.java` line 47 at r8](https://reviewable.io/reviews/harness/harness-core/39191#-NGboa6N1UsSzdLg865s:-NGboa6N1UsSzdLg865t:bwtej80) ([raw file](https://github.com/harness/harness-core/blob/edae3242cbbaeeb1fb2e2b2017f90260e8501bad/960-persistence/src/test/java/io/harness/iterator/IteratorExecutionHandlerTest.java#L47)):*
> ```Java
>   public void testIteratorEnabled() {
>     // Start the iterator with dummy config
>     IteratorExecutionHandler.DynamicIteratorConfigOption iteratorConfigOption =
> ```

// Arrange 
IteratorExecutionHandler.DynamicIteratorConfigOption.builder iteratorConfigOption = buildDefaultConfiguration(TEST_ITERATOR_NAME).toBuilder();
iteratorConfigOption.enabled(true);

// Act
 iteratorExecutionHandler.startIterator(iteratorConfigOption);

// Assert
assertThat(iteratorHandler.getIteratorName()).isEqualTo(TEST_ITERATOR_NAME);



_Code quote:_
```Java
 IteratorExecutionHandler.DynamicIteratorConfigOption iteratorConfigOption =
```

gn-harness (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 46 file(s) to review. 1 change requests. 7. unresolved discussions. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandler.java` line 15 at r8](https://reviewable.io/reviews/harness/harness-core/39191#-NGbplUJ7hgQLPJqbbt7:-NGbplUJ7hgQLPJqbbt8:bqjbiz8) ([raw file](https://github.com/harness/harness-core/blob/edae3242cbbaeeb1fb2e2b2017f90260e8501bad/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandler.java#L15)):*
> ```Java
> 
> @Slf4j
> public class IteratorExecutionHandler {
> ```

Create three entities here:

IteratorExecutionHandler - interface
IteratorExecutionHandlerImpl - class implements IteratorExecutionHandler
IteratorConfigWatcher.

IteratorExecutionHandlerImpl should have a reference to IteratorConfigWatcher - this class would start the watcher (probably with a function call in the wingsApplication like startConfigWatcher)

IteratorConfigWatcher should have a reference to IteratorExecutionHandler. i.e. when IteratorConfigWatcher is created, it's constructor should have IteratorConfigWatcher as an argument, with actual construction looking something like:

// IteratorExecutionHandlerImpl.java
IteratorConfigWatcher watcher = new IteratorConfigWatcher(this);

This interface approach would help avoid cycles between Impl and watcher.



_Code quote:_
```Java
public class IteratorExecutionHandler
```

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 49 file(s) to review. 1 change requests. 9. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 31 at r8](https://reviewable.io/reviews/harness/harness-core/39191#-NGbULOa-10OwrUMToDr:-NGgd9K00H8jrFdakBIY:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/edae3242cbbaeeb1fb2e2b2017f90260e8501bad/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L31)):*

nit: registerIterator*
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandler.java` line 15 at r8](https://reviewable.io/reviews/harness/harness-core/39191#-NGbplUJ7hgQLPJqbbt7:-NGgdAnT6n5hmPAG3Ros:bo5c5oe) ([raw file](https://github.com/harness/harness-core/blob/edae3242cbbaeeb1fb2e2b2017f90260e8501bad/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandler.java#L15)):*

Create three entities here:

IteratorExecutionHandler - interface
IteratorExecutionHandlerImpl - class implements IteratorExecutionHandler
IteratorConfigWatcher.

IteratorExecutionHandlerImpl should have a reference to IteratorConfigWatcher - this class would start the watcher (probably with a function call in the wingsApplication like startConfigWatcher)

IteratorConfigWatcher should have a reference to IteratorExecutionHandler. i.e. when IteratorConfigWatcher is created, it's constructor should have IteratorConfigWatcher as an argument, with actual construction looking something like:

// IteratorExecutionHandlerImpl.java
IteratorConfigWatcher watcher = new IteratorConfigWatcher(this);

This interface approach would help avoid cycles between Impl and watcher.
</blockquote></details>

Done. Thanks for the inputs here :). 
___
*[`960-persistence/src/test/java/io/harness/iterator/IteratorExecutionHandlerTest.java` line 47 at r8](https://reviewable.io/reviews/harness/harness-core/39191#-NGboa6N1UsSzdLg865s:-NGgd-JdConLwrwTfaeF:b-b2zxjc) ([raw file](https://github.com/harness/harness-core/blob/edae3242cbbaeeb1fb2e2b2017f90260e8501bad/960-persistence/src/test/java/io/harness/iterator/IteratorExecutionHandlerTest.java#L47)):*

// Arrange 
IteratorExecutionHandler.DynamicIteratorConfigOption.builder iteratorConfigOption = buildDefaultConfiguration(TEST_ITERATOR_NAME).toBuilder();
iteratorConfigOption.enabled(true);

// Act
 iteratorExecutionHandler.startIterator(iteratorConfigOption);

// Assert
assertThat(iteratorHandler.getIteratorName()).isEqualTo(TEST_ITERATOR_NAME);
</blockquote></details>

Done. Thanks for the inputs on organizing unit tests :)

Raghav-Murali:
trigger feature-build

Raghav-Murali:
trigger feature-build

Raghav-Murali:
trigger feature-build

gn-harness (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 48 file(s) to review. 1 change requests. 6. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/io/harness/ccm/license/CeLicenseExpiryHandler.java` line 41 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGohWmB0TdYBsqy7Tf6:-NGohWmB0TdYBsqy7Tf7:bmeoz00) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/400-rest/src/main/java/io/harness/ccm/license/CeLicenseExpiryHandler.java#L41)):*
> ```Java
>   @Override
>   protected void createAndStartIterator(
>       PersistenceIteratorFactory.PumpExecutorOptions executorOptions, Duration targetInterval) {
> ```

nit: In a future PR, we can look into combining these two arguments -- and may be just pass the configuration directly



_Code quote:_
```Java
PersistenceIteratorFactory.PumpExecutorOptions executorOptions, Duration targetInterval
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigWatcher.java` line 74 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGoVZtpEPy-0h3LFf3X:-NGoVZtpEPy-0h3LFf3Y:bg23dt0) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorConfigWatcher.java#L74)):*
> ```Java
>           objectMapper.readValue(dir.toFile(), IteratorExecutionHandler.DynamicIteratorConfigOption[].class);
>     } catch (IOException e) {
>       log.error("Received exception while reading configMap", e);
> ```

nit: iteratorConfigFile - please add this to exception



_Code quote:_
```Java
configMap
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandler.java` line 14 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGoXl4Q4vfuiN3pITnr:-NGoXl4Q4vfuiN3pITns:b-53eggg) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandler.java#L14)):*
> ```Java
>   @Value
>   @Builder
>   class DynamicIteratorConfigOption {
> ```

nit: DynamicIteratorConfiguration or DynamicIteratorConfig?
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 96 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGoX71n2wuwrZ13n-YM:-NGoX71n2wuwrZ13n-YN:by9sdo5) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L96)):*
> ```Java
>    */
>   @Override
>   public void applyConfiguration(IteratorExecutionHandler.DynamicIteratorConfigOption iteratorConfigOption) {
> ```

nit: remove prefix or stay consistent across this file



_Code quote:_
```Java
IteratorExecutionHandler.
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 116 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGocTz71EpgM1lFp4fB:-NGocTz71EpgM1lFp4fC:b-vgrnb2) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L116)):*
> ```Java
>       // The iterator has only been registered and configuration is not yet applied.
>       applyIteratorConfig(iteratorConfigOption);
>     } else {
> ```

Why do we need to handle INIT as a special case?



_Code quote:_
```Java
    if (iterator.iteratorStateValue == IteratorStateValues.INIT) {
      // The iterator has only been registered and configuration is not yet applied.
      applyIteratorConfig(iteratorConfigOption);
    } else {
```
___
*[`960-persistence/src/test/java/io/harness/iterator/IteratorExecutionHandlerImplTest.java` line 131 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGodcrd33x6uO5PxGVA:-NGodcrd33x6uO5PxGVB:b-ivz6g2) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/test/java/io/harness/iterator/IteratorExecutionHandlerImplTest.java#L131)):*
> ```Java
>         .isEqualTo(IteratorExecutionHandlerImpl.IteratorStateValues.RUNNING);
>     assertThat(iteratorState.getIteratorConfigOption()).isEqualTo(iteratorConfigOptionList.get(0));
>     assertThat(iteratorBaseHandler.getIterator().getExecutorService().isShutdown()).isEqualTo(false);
> ```

this is all arrange - you dont need the first batch of asserts here. there is a separate test to verify enable iterator case.

and same for other test cases..

In general, a test only tests one thing - with one Arrange/Act and assert



_Code quote:_
```Java
    // Arrange
    String iteratorName = testConfigList.get(0).getName();
    IteratorBaseHandler iteratorBaseHandler = new IteratorTestHandler(iteratorName);
    List<IteratorExecutionHandler.DynamicIteratorConfigOption> iteratorConfigOptionList =
        new ArrayList<>(Arrays.asList(testConfigList.get(0)));
    writeTestIteratorConfigToFile(iteratorConfigOptionList);
    iteratorBaseHandler.registerIterator(iteratorExecutionHandler);

    // Act
    iteratorExecutionHandler.startIterators();

    // Assert
    assertThat(iteratorExecutionHandler.getIteratorState().containsKey(iteratorName)).isEqualTo(true);
    IteratorExecutionHandlerImpl.IteratorState iteratorState =
        iteratorExecutionHandler.getIteratorState().get(iteratorName);
    assertThat(iteratorState.getIteratorStateValue())
        .isEqualTo(IteratorExecutionHandlerImpl.IteratorStateValues.RUNNING);
    assertThat(iteratorState.getIteratorConfigOption()).isEqualTo(iteratorConfigOptionList.get(0));
    assertThat(iteratorBaseHandler.getIterator().getExecutorService().isShutdown()).isEqualTo(false);
```
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorLoopModeHandler.java` line 76 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGodF78ADpph4V4jfYc:-NGodF78ADpph4V4jfYd:bciapv0) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorLoopModeHandler.java#L76)):*
> ```Java
>   @Override
>   protected abstract void createAndStartIterator(
>       PersistenceIteratorFactory.PumpExecutorOptions executorOptions, Duration targetInterval);
> ```

Why *PumpExecutorOptions* Are there in the loop mode handler?



_Code quote:_
```Java
PumpExecutorOptions
```

gn-harness (dismissed):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 48 file(s) to review. 2. unresolved discussions. 1 of 2 approvals obtained.

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 48 file(s) to review. 5. unresolved discussions. 1 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/io/harness/ccm/license/CeLicenseExpiryHandler.java` line 41 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGohWmB0TdYBsqy7Tf6:-NGpF32P7M2AL_YIC4ye:b-y967z4) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/400-rest/src/main/java/io/harness/ccm/license/CeLicenseExpiryHandler.java#L41)):*

nit: In a future PR, we can look into combining these two arguments -- and may be just pass the configuration directly
</blockquote></details>

Done. Created this ticket to follow-up on it PL-29757
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorConfigWatcher.java` line 74 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGoVZtpEPy-0h3LFf3X:-NGpGPDi1-bri3Y2eMoe:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorConfigWatcher.java#L74)):*

nit: iteratorConfigFile - please add this to exception
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandler.java` line 14 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGoXl4Q4vfuiN3pITnr:-NGpGk2nAIuM-HwAr3yh:b-qwb6g) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandler.java#L14)):*

nit: DynamicIteratorConfiguration or DynamicIteratorConfig?
</blockquote></details>

Done. Updated to DynamicIteratorConfig
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 96 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGoX71n2wuwrZ13n-YM:-NGpHFTF4QVqu7Q0jhFH:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L96)):*

nit: remove prefix or stay consistent across this file
</blockquote></details>

Done.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 116 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGocTz71EpgM1lFp4fB:-NGpHMfBBGb9VNWcFzPA:bd0nzq9) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L116)):*

Why do we need to handle INIT as a special case?
</blockquote></details>

It might be possible that an iterator is registered using `registerIterators()` but the initial configMap or file doesn't contain any configuration for the iterator. In such cases the iterator is only registered but no config is applied to it yet. Thus, INIT means iterator is registered but it doesn't have any configuration yet. So when an iterator is in INIT state we will just apply the received configuration and do not compare it with existing configuration as there is none. 

I was thinking on whether to use INIT or REGISTERED as the name for this state but using REGISTERED can cause confusion since if it goes to else part people might think the iterator is not registered. Thus, used name as INIT.
___
*[`960-persistence/src/test/java/io/harness/iterator/IteratorExecutionHandlerImplTest.java` line 131 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGodcrd33x6uO5PxGVA:-NGpTVul5PXFhkvIYQ7O:bpujwlm) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/test/java/io/harness/iterator/IteratorExecutionHandlerImplTest.java#L131)):*

this is all arrange - you dont need the first batch of asserts here. there is a separate test to verify enable iterator case.

and same for other test cases..

In general, a test only tests one thing - with one Arrange/Act and assert
</blockquote></details>

Done, got rid of the redundant asserts.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorLoopModeHandler.java` line 76 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGodF78ADpph4V4jfYc:-NGpJvr59dUJkt8dR3Gv:b-qn13u2) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorLoopModeHandler.java#L76)):*

Why *PumpExecutorOptions* Are there in the loop mode handler?
</blockquote></details>

Actually this contains the variables needed for an executorService - poolSize, interval and threadPool name. However, this was being used in other services like Platform and Pipeline to create Loop iterators. I was skeptical of renaming it in this PR as it will add even more files spanning other services that are not relevant to configuring iterators dynamically. I will rename it as part of this JIRA PL-29757

Raghav-Murali:
trigger feature-build

Raghav-Murali:
trigger codeformat

Raghav-Murali:
trigger feature-build

archit-harness (approved):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 48 file(s) to review. 5. unresolved discussions. 1 of 2 approvals obtained.

sahilhindwani-harness (approved):
> Reviewed 11 of 56 files at r1, 3 of 8 files at r4, 4 of 48 files at r7, 41 of 47 files at r9, 1 of 1 files at r12, 5 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 46 file(s) to review. 5. unresolved discussions.

gn-harness (approved):
> Reviewed 11 of 56 files at r1, 3 of 8 files at r4, 4 of 48 files at r7, 41 of 47 files at r9, 1 of 1 files at r12, 5 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 4. unresolved discussions. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 116 at r11](https://reviewable.io/reviews/harness/harness-core/39191#-NGocTz71EpgM1lFp4fB:-NGw6T8B05cTetuacmu0:b-uon4n4) ([raw file](https://github.com/harness/harness-core/blob/ac7a4e6945cb946c14608bec91fd37308ab4ee9d/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L116)):*

It might be possible that an iterator is registered using `registerIterators()` but the initial configMap or file doesn't contain any configuration for the iterator. In such cases the iterator is only registered but no config is applied to it yet. Thus, INIT means iterator is registered but it doesn't have any configuration yet. So when an iterator is in INIT state we will just apply the received configuration and do not compare it with existing configuration as there is none. 

I was thinking on whether to use INIT or REGISTERED as the name for this state but using REGISTERED can cause confusion since if it goes to else part people might think the iterator is not registered. Thus, used name as INIT.
</blockquote></details>

Naming is ok. 

My concern was that INIT does not seem to add much additional value, especially in this function. Would something like the below works? (no need to be blocked on this, feel free to submit it as is)


// SKIP the check on the iterator state. If config is there, compare it and return early. You can remove all this nesting logic.
====

 if (iterator.iteratorConfigOption != NULL && iteratorConfigOption.equals(iterator.iteratorConfigOption)) {
   // No change in the iterator config, thus ignore.
   return;
  }

  int configAttempts = 3;
  for (int attempt = 0; attempt < configAttempts; attempt++) {
      if (applyIteratorConfig(iteratorConfigOption)) {
        // Configuration applied successfully, so return
         return;
       }
       sleep(ofMillis(1000));
   }
====

raghuAtWings (approved):
> Reviewed 3 of 8 files at r4, 4 of 48 files at r7, 41 of 47 files at r9, 1 of 1 files at r12, 5 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: 2. unresolved discussions.

gn-harness (approved):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/39191)* status: :shipit: complete! ✓

PR #39809, opened 2022-11-08
feat: [SPG-482]: Change to measure postProcess instead processStencil

Fernando-Dourado:
## Describe your changes
- Cleaning up log events for prod stability


Fernando-Dourado:
trigger ti3

PR #40826, opened 2022-11-29
fix: [DEL-5460]: Add included kubectl for immutable-delegate to system $PATH variable

raghuAtWings:
## Describe your changes
1. change the base image of immutable delegate to redhat/ubi8-minimal:8.7
2. set the PATH to include v1.19.2 kubectl baked in the image

## Checklist

## Comment Triggers
<details>

</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
- SonarScan: `trigger ss`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/40826)

### DIFF
--- a/dockerization/delegate/Dockerfile-immutable
+++ b/dockerization/delegate/Dockerfile-immutable
@@ -1,4 +1,4 @@
-FROM redhat/ubi8-minimal:latest
+FROM redhat/ubi8-minimal:8.7
 
 LABEL name="harness/delegate-immutable" \
       vendor="Harness" \
@@ -36,10 +36,9 @@ ENV HOME=/opt/harness-delegate
 ENV CLIENT_TOOLS_DOWNLOAD_DISABLED=$disable_client_tools
 ENV INSTALL_CLIENT_TOOLS_IN_BACKGROUND=$disable_client_tools
 ENV JAVA_HOME=/opt/java/openjdk/
-ENV PATH="$JAVA_HOME/bin:${PATH}"
+ENV PATH="$JAVA_HOME/bin:/opt/harness-delegate/client-tools/kubectl/v1.19.2/:${PATH}"
 
 COPY ./delegate.jar ./delegate.jar
-
 USER 1001
 
 CMD [ "./start.sh" ]


xingchi-jin (approved):
LG for this particular issue.

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40826#-:-NI47ZfH-3gEG0oMWFcU:bwyyzcl)* status: :shipit: complete! ✓

dockerization/delegate/Dockerfile-immutable
```diff
@@ -36,10 +36,9 @@ ENV HOME=/opt/harness-delegate
 ENV CLIENT_TOOLS_DOWNLOAD_DISABLED=$disable_client_tools
 ENV INSTALL_CLIENT_TOOLS_IN_BACKGROUND=$disable_client_tools
 ENV JAVA_HOME=/opt/java/openjdk/
-ENV PATH="$JAVA_HOME/bin:${PATH}"
+ENV PATH="$JAVA_HOME/bin:/opt/harness-delegate/client-tools/kubectl/v1.19.2/:${PATH}"
```
mdmilic, line 39: I think this is a very bad idea. I understand some customers requested it but we have a very simple one liner (in init script) which they can do in their environment to get it. 
It was a mistake to put this in legacy delegates as well and we can't get rid of it now, so would be great we avoid the same mistake. Reasons why it's bad:
* You are tied to a version. You can't change this. If you change version you can will break customers (IIRC something like this happened with legacy when we tried to remove this)
* It will affect existing customers with and without custom image. This will now behave as if they've put custom kubectl binary (and will now never use other version even if tasks request it), and for custom images it could even override their custom binary, depending on how they've configured it (e.g. if they've did $PATH:/my/path )
* Still not same as legacy - there it's latest kubectl
* Applies to minimal image as well which doesn't have the tool. I think it will not break it right now though, but not future proof.

mdmilic (changes requested):

raghuAtWings:
*[`dockerization/delegate/Dockerfile-immutable` line 39 at r1](https://reviewable.io/reviews/harness/harness-core/40826#-NI45DLWYDEC6qJzvOuW-r1-39:-NI53Ns1A957Nu3EX_3u:b-6ujbh9) ([raw file](https://github.com/harness/harness-core/blob/e35f0213eccc9a75c57c34a724ac4c314aa79b53/dockerization/delegate/Dockerfile-immutable#L39)):*

I think this is a very bad idea. I understand some customers requested it but we have a very simple one liner (in init script) which they can do in their environment to get it. 
It was a mistake to put this in legacy delegates as well and we can't get rid of it now, so would be great we avoid the same mistake. Reasons why it's bad:
* You are tied to a version. You can't change this. If you change version you can will break customers (IIRC something like this happened with legacy when we tried to remove this)
* It will affect existing customers with and without custom image. This will now behave as if they've put custom kubectl binary (and will now never use other version even if tasks request it), and for custom images it could even override their custom binary, depending on how they've configured it (e.g. if they've did $PATH:/my/path )
* Still not same as legacy - there it's latest kubectl
* Applies to minimal image as well which doesn't have the tool. I think it will not break it right now though, but not future proof.

</blockquote></details>

should we put this in our default generated INIT_SCRIPT?

xingchi-jin:
*[`dockerization/delegate/Dockerfile-immutable` line 39 at r1](https://reviewable.io/reviews/harness/harness-core/40826#-NI45DLWYDEC6qJzvOuW-r1-39:-NI5CtaN5nNJkxlJ3Hs4:b-wfmexx) ([raw file](https://github.com/harness/harness-core/blob/e35f0213eccc9a75c57c34a724ac4c314aa79b53/dockerization/delegate/Dockerfile-immutable#L39)):*

should we put this in our default generated INIT_SCRIPT? 
</blockquote></details>

@mdmilic 's points are inspiring.
I think changing PATH will affect existing customers is the risk here.

Then the question in my mind is why do we provide tools to them?  If customers treat delegate as a blank execution sandbox to run shell commands or scripts, why don't they install kubectl in their script.

mdmilic:
> should we put this in our default generated INIT_SCRIPT

No, this would have the same effect as putting it in Dockerfile. I don't think we should do this for now. After we cleanup to have one client tool version, might be feasible. 
At that point I think this would be the only impact:
* For custom images it could even override their custom binary, depending on how they've configured it (e.g. if they've did $PATH:/my/path )

PR #40871, opened 2022-11-30
feat: [DEL-5371]: have delegate tasks count limit per account

arpit3107:
## Describe your changes

Earlier:
Manager was having a configurable limit to as of how many delegate tasks (per rank) can be created within 1 min. There are 3 possible ranks associated with a task. 

So we wanted to have this task creation limit to be per account, instead of global. Also we want to make this limit configurable per account (in future) after having the metrics of per account usage. 

To start with we are keeping this limit as hardcoded constant and can tune it in future. Also we are not bothering with ranks, as task owners dont really care about this rank.

Tests:
1. Delegate task metrics are visible in gcp monitoring.
2. Created tasks and got the metrics.
3. Exisiting delegate tasks metrics are working fine.
4. In local, i set this limit as 1 and verified that no new tasks was getting created.

## Checklist
- [x] If hashcheck failed I followed [the checklist](https://harness.atlassian.net/wiki/spaces/DEL/pages/21016838831/PR+Codebasehash+Check+merge+checklist) and updated this PR description with my answers to make sure I'm not introducing any breaking changes.

## Comment Triggers
<details>

</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
- SonarScan: `trigger ss`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)


mdmilic (changes requested):
I think we should either not throttle or give priority (RANK :) ) to sync tasks as they are usually related to UI operation, so this could lead to UI slowness. It is however probably OK to do this in next iteration, but just bear it in mind or JIRA :). 

> Reviewed 7 of 7 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871#-:-NI8Ee9y0v-iOBMWmCa5:bf8vpbd)* status: 1 change requests. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 1836 at r1](https://reviewable.io/reviews/harness/harness-core/40871#-NI8Ga_-9Qj6xstrr5P2:-NI8Ga_-9Qj6xstrr5P3:b46vcim) ([raw file](https://github.com/harness/harness-core/blob/be81442f72ed6883af9c0bee2b62d664409ecc72/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L1836)):*
> ```Java
> 
>   @Override
>   public long delegateTasksCount(@NotNull String accountId) {
> ```

Circular dependency with cache helper. Move it to cache helper. 
___
*[`400-rest/src/main/resources/metrics/metricDefinitions/delegate_task_metrics.yaml` line 45 at r1](https://reviewable.io/reviews/harness/harness-core/40871#-NI8G9XwEAH9woQU3iZh:-NI8G9XwEAH9woQU3iZi:b-ynwjrp) ([raw file](https://github.com/harness/harness-core/blob/be81442f72ed6883af9c0bee2b62d664409ecc72/400-rest/src/main/resources/metrics/metricDefinitions/delegate_task_metrics.yaml#L45)):*
> ```YAML
>     type: Count
>     unit: "1"
>   - metricName: delegate_task_count_per_account
> ```

lets call it just `delegate_task_count` aggregations are done on higher level, so in grafana or metric explorer you decide if this wants to be per account or per node or global etc.. and just aggregate accordingly
___
*[`420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java` line 21 at r1](https://reviewable.io/reviews/harness/harness-core/40871#-NI8LsRz9axoJcKauSh7:-NI8LsRz9axoJcKauSh8:b-ilyvfc) ([raw file](https://github.com/harness/harness-core/blob/be81442f72ed6883af9c0bee2b62d664409ecc72/420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java#L21)):*
> ```Java
> @OwnedBy(DEL)
> @Singleton
> public class DelegateTasksCountCacheHelper {
> ```

why helper. Lets just call it cache.
___
*[`420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java` line 44 at r1](https://reviewable.io/reviews/harness/harness-core/40871#-NI8LgNI3R-7BJa9hcAg:-NI8LgNK-TO8apJvapoP:b593d76) ([raw file](https://github.com/harness/harness-core/blob/be81442f72ed6883af9c0bee2b62d664409ecc72/420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java#L44)):*
> ```Java
>   }
> 
>   public void recordDelegateTasksCountMetric() {
> ```

I think this method should be in metrics publisher to avoid bringing in metricsService as a dependency. 
___
*[`420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsServiceImpl.java` line 130 at r1](https://reviewable.io/reviews/harness/harness-core/40871#-NI8MGiHCGlTbdJ9-sw4:-NI8MGiHCGlTbdJ9-sw5:brrqug0) ([raw file](https://github.com/harness/harness-core/blob/be81442f72ed6883af9c0bee2b62d664409ecc72/420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsServiceImpl.java#L130)):*
> ```Java
> 
>   @Override
>   public void recordDelegateTaskCountMetricsPerAccount(String accountId, String metricName, long value) {
> ```

You should either rename method to something more generic like `recordDelegateMetricsPerAccount` or don't need to provide metricName as input since the name already suggests which metric it's about.

jennyjames21:
*[`420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java` line 24 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NI8nnES3BpEtdzuDqDz:-NI8nnES3BpEtdzuDqE-:b-cxijrt) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java#L24)):*
> ```Java
> @Singleton
> @Slf4j
> public class DelegateTasksCountCacheHelper {
> ```

Do we need a separate class for this? Can we combine to DelegateCache where we track all delegate related cache?  DelegateCache is already injected in DelegateTaskClassicServceImpl. It good to keep all delegate-related cache in one class.

Or create a new service for delegateTaskCacheService.

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 361 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NI8pepZ9ohSR1PXEQ-Y:-NI8pepZ9ohSR1PXEQ-Z:b-r2v83f) ([raw file](https://github.com/harness/harness-core/blob/a81ea0b81aefa13325f958d92d44903ab1f32b5f/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L361)):*
> ```Java
>         return mainConfiguration.getPortal().getCriticalDelegateTaskRejectAtLimit();
>       default:
>         throw new InvalidArgumentsException("Unsupported delegate task rank level " + rank);
> ```

So we removing ranking system here? Parked task comes with optional category which can come from multiple accounts and overwhelm system.

420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java
```diff
@@ -0,0 +1,60 @@
...
+    Long delegateTasksCount = delegateTasksCountCache.getIfPresent(accountId);
+    if (delegateTasksCount == null) {
+      delegateTasksCount = delegateTasksCount(accountId);
+      delegateTasksCountCache.put(accountId, delegateTasksCount);
+    }
+    return delegateTasksCount;
+  }
+
+  public void recordDelegateTasksCountMetric() {
+    log.info("Arpit: recording size {}", delegateTasksCountCache.estimatedSize());
```
abhinav-harness, line 48: arpit in logs

gaurav-jain1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: 2 change requests. 1. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 339 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGS8Rq178TgghPNdHA:-NIGS8Rq178TgghPNdHB:b-6x8id9) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L339)):*
> ```Java
>     long currentTaskCountOfAccount = delegateTasksCountCacheHelper.getDelegateTasksCount(accountId);
>     if (currentTaskCountOfAccount > MAX_DELEGATE_TASKS_COUNT_PER_ACCOUNT) {
>       throw new RateLimitExceededException("Rate limit reached for tasks for account " + accountId);
> ```

can you also attach the threshold value and current task count as well.



_Code quote:_
```Java
Rate limit reached for tasks for account 
```
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 584 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGSTbBA0TdljqbrzL3:-NIGSTbBA0TdljqbrzL4:bmgj9ne) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L584)):*
> ```Java
>           }
>         }
>         checkTaskCountLimitPerAccount(task.getAccountId());
> ```

We should check in very stages of ProcessDelegateTask and skip all of capabilities check if limit itself has crossed.



_Code quote:_
```Java
checkTaskCountLimitPerAccount(task.getAccountId());
```
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 701 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGSeyX6JPi7BLsLWZq:-NIGSeyX6JPi7BLsLWZr:b-2guby3) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L701)):*
> ```Java
>         }
> 
>         checkTaskCountLimitPerAccount(task.getAccountId());
> ```

same here



_Code quote:_
```Java
checkTaskCountLimitPerAccount(task.getAccountId());
```
___
*[`420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java` line 24 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NI8nnES3BpEtdzuDqDz:-NIGTMolDvcwTiHk_Chq:beogrfl) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java#L24)):*
> r create a new service for delegateTaskCacheService.
this is a good suggestion, we have lot many caches and to use all of them wisely, make sense to make a one unified cache. May be we can take that up as a separate project @jennyjames21 
___
*[`420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java` line 49 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGU_qb4IK9oPHAOYOP:-NIGU_qc4z0ni1VDHeY8:b-u98h06) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java#L49)):*
> ```Java
>   public void recordDelegateTasksCountMetric() {
>     log.info("Arpit: recording size {}", delegateTasksCountCache.estimatedSize());
>     for (Map.Entry<String, Long> entry : delegateTasksCountCache.asMap().entrySet()) {
> ```

will iterating over this map, incurs some cache penalty which can increase cache lookup time ?



_Code quote:_
```Java
 (Map.Entry<String, Long> entry : delegateTasksCountCache.asMap().entrySe
```

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 584 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGSTbBA0TdljqbrzL3:-NIGXNts22iZTpm5ViMk:b-mylqsc) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L584)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

We should check in very stages of ProcessDelegateTask and skip all of capabilities check if limit itself has crossed.
</blockquote></details>

Reaching task limit is a very very rare case(we may see less 5 times in prod last year), adding checks everywhere will create unnecessary code for every task we process. Once per task is good enough, you can change the order where to keep that check. 
I would avoid changing that order in order, Arpit you can make that call

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 701 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGSeyX6JPi7BLsLWZq:-NIGY_IV25vYLpTRApp3:b-pdgkdy) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L701)):*
> Reaching task limit is a very very rare case(we may see less 5 times in prod last year), adding checks everywhere will create unnecessary code for every task we process. Once per task is good enough, you can change the order where to keep that check.
> I would avoid changing that order in order, Arpit you can make that call

gaurav-jain1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: 10 file(s) to review. 2 change requests. 3. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 584 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGSTbBA0TdljqbrzL3:-NIHOCPd6JyLikBCye2n:bam0o1v) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L584)):*

Reaching task limit is a very very rare case(we may see less 5 times in prod last year), adding checks everywhere will create unnecessary code for every task we process. Once per task is good enough, you can change the order where to keep that check. 
I would avoid changing that order in order, Arpit you can make that call
</blockquote></details>

I mistyped my earlier comment, I meant to move this check to beginning of this method, to avoid operations if we reach out limit.
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 354 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIHNqC4AWFSOT55eDvt:-NIHNqC51bsSS7MSdz6S:bushan9) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L354)):*
> ```Java
>       case IMPORTANT:
>         return delegateCache.getImportantDelegateTasksCount(accountId) > MAX_IMPORTANT_DELEGATE_TASKS_COUNT_PER_ACCOUNT;
>       default:
> ```

How about critical ? do you want to handle in the same range ?



_Code quote:_
```Java
default:
```
___
*[`420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java` line 143 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIHP4Ce3F9a5BGH5Iol:-NIHP4Ce3F9a5BGH5Iom:b-1jsrhu) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java#L143)):*
> ```Java
> 
>   private void recordDelegateTasksByRankMetrics() {
>     log.info("Arpit: recording delegate tasks count metrics. OptionalSize {} ImportantSize {}",
> ```

Remove your name :)



_Code quote:_
```Java
Arpit:
```
___
*[`420-delegate-service/src/main/java/io/harness/service/impl/DelegateCacheImpl.java` line 70 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIHObmWEn_Pu7wyuYuO:-NIHObmWEn_Pu7wyuYuP:b-rjmujy) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/420-delegate-service/src/main/java/io/harness/service/impl/DelegateCacheImpl.java#L70)):*
> ```Java
>           });
> 
>   private LoadingCache<ImmutablePair<String, String>, DelegateGroup> delegateGroupCache =
> ```

I was suggesting to unify all these different caches to one so that we can utilize them wisely. Not something to take care in this PR, but something we can pick up later.

arpit3107 (commented):
That task owners can decide themselves.

> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871#-:-NIVvvbG61WAkR7BVdEq:byc70pl)* status: 10 file(s) to review. 2 change requests. 9. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 339 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGS8Rq178TgghPNdHA:-NIVw6MtCOLTo6WvpVOB:bb7oryy) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L339)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

can you also attach the threshold value and current task count as well.
</blockquote></details>

sure
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 361 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NI8pepZ9ohSR1PXEQ-Y:-NIVwNhSBsJXfMCj_vmx:bycoomy) ([raw file](https://github.com/harness/harness-core/blob/a81ea0b81aefa13325f958d92d44903ab1f32b5f/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L361)):*

So we removing ranking system here? Parked task comes with optional category which can come from multiple accounts and overwhelm system.
</blockquote></details>

Optional and Important ranks will be there
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 584 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGSTbBA0TdljqbrzL3:-NIVwTGb-olyLcjm3x_T:bgdddko) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L584)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

I mistyped my earlier comment, I meant to move this check to beginning of this method, to avoid operations if we reach out limit.
</blockquote></details>

We can first get the metrics and then decide on this. Till that time the check can be on earlier level only
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 701 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGSeyX6JPi7BLsLWZq:-NIVwf489DeQqROpJaaQ:b-x1g8lo) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L701)):*
> We can first get the metrics and then decide on this. Till that time the check can be on earlier level only
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 354 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIHNqC4AWFSOT55eDvt:-NIVw7NA9IvhRKbmkUnP:be8lujk) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L354)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

How about critical ? do you want to handle in the same range ?
</blockquote></details>

Critical is deprecated. Optional and important are good enough. Actually noone is caring about ranks while creating a task. Only CI team uses optional, but no one uses other two
___
*[`400-rest/src/main/resources/metrics/metricDefinitions/delegate_task_metrics.yaml` line 45 at r1](https://reviewable.io/reviews/harness/harness-core/40871#-NI8G9XwEAH9woQU3iZh:-NIVwlvh5tf4dwiklzLA:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/be81442f72ed6883af9c0bee2b62d664409ecc72/400-rest/src/main/resources/metrics/metricDefinitions/delegate_task_metrics.yaml#L45)):*

lets call it just `delegate_task_count` aggregations are done on higher level, so in grafana or metric explorer you decide if this wants to be per account or per node or global etc.. and just aggregate accordingly
</blockquote></details>

Done.
___
*[`420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java` line 143 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIHP4Ce3F9a5BGH5Iol:-NIVwo7F5KBuhZ3MiB_w:bnp4nxx) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java#L143)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

Remove your name :)
</blockquote></details>

This is for pr testing. I will remove before merging pr
___
*[`420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsServiceImpl.java` line 130 at r1](https://reviewable.io/reviews/harness/harness-core/40871#-NI8MGiHCGlTbdJ9-sw4:-NIVwuGh68BZn6eO9iw_:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/be81442f72ed6883af9c0bee2b62d664409ecc72/420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsServiceImpl.java#L130)):*

You should either rename method to something more generic like `recordDelegateMetricsPerAccount` or don't need to provide metricName as input since the name already suggests which metric it's about.
</blockquote></details>

Done.

gaurav-jain1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: Some of the checks are too old. 10 file(s) to review. 2 change requests. 9. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/test/java/software/wings/service/delegate/DelegateTaskServiceClassicTest.java` line 368 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIW_j5k79WANU4f59Qx:-NIW_j5lEgpFzLBJAtbM:b-oe771u) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/400-rest/src/test/java/software/wings/service/delegate/DelegateTaskServiceClassicTest.java#L368)):*
> ```Java
>                                     .uuid(generateUuid())
>                                     .accountId(ACCOUNT_ID)
>                                     //.rank(DelegateTaskRank.IMPORTANT)
> ```

Remove if not needed



_Code quote:_
```Java
 //.rank(DelegateTaskRank.IMPORTANT)
```
___
*[`420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java` line 148 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIW_zUL1E1xNTh9zRUu:-NIW_zUL1E1xNTh9zRUv:b-9st5ti) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java#L148)):*
> ```Java
>     try {
>       for (Map.Entry<String, Long> entry : delegateCache.getMapOfOptionalDelegateTasksCount().entrySet()) {
>         log.info("Arpit: optional key {} value {}", entry.getKey(), entry.getValue());
> ```

remove this log.info before merging your pr



_Code quote:_
```Java
log.info("Arpit: optional 
```
___
*[`950-delegate-tasks-beans/src/main/java/io/harness/delegate/beans/DelegateTaskRank.java` line 19 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIWaP7TBxapjq4hlbYG:-NIWaP7TBxapjq4hlbYH:bdejrgt) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/950-delegate-tasks-beans/src/main/java/io/harness/delegate/beans/DelegateTaskRank.java#L19)):*
> ```Java
> public enum DelegateTaskRank {
>   OPTIONAL,
>   @Deprecated IMPORTANT, // instead use the rank as NULL
> ```

Is this rank also deprecated ? If yes, why were tracking it separately than ?



_Code quote:_
```Java
@Deprecated IMPORTANT,
```

arpit3107 (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: 13 file(s) to review. 2 change requests. 13. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/resources/metrics/metricDefinitions/delegate_task_metrics.yaml` line 45 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIcU-zaAnh1vf3G9ZRc:-NIl0UDW1MGJtgm16OwQ:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/400-rest/src/main/resources/metrics/metricDefinitions/delegate_task_metrics.yaml#L45)):*

I don't think these should be two metrics, it should be one metric, but then you can add additional label related to rank if you need to. Again this is about aggregation and filtering which is done on much higher level and this way will allow you to choose if you want number of tasks per account, or per account and rank etc.. Data/cost wise it would produce about the same amount of data (I think depending on the frequency of one rank vs the other)
</blockquote></details>

Done.
___
*[`420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java` line 148 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIW_zUL1E1xNTh9zRUu:-NIl4RRyCWHs3l_kKlUP:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java#L148)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

remove this log.info before merging your pr
</blockquote></details>

Done.
___
*[`950-delegate-tasks-beans/src/main/java/io/harness/delegate/beans/DelegateTaskRank.java` line 19 at r4](https://reviewable.io/reviews/harness/harness-core/40871#-NIWaP7TBxapjq4hlbYG:-NIl4TmDDfkiVI2hIePZ:b-3zusuq) ([raw file](https://github.com/harness/harness-core/blob/48dcd999e21cfb6d3443b9e67b1800aca1431179/950-delegate-tasks-beans/src/main/java/io/harness/delegate/beans/DelegateTaskRank.java#L19)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

Is this rank also deprecated ? If yes, why were tracking it separately than ?
</blockquote></details>

Removed deprecated label, tasks having rank NULL will be marked as IMPORTANT while processing task
___
*[`420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java` line 21 at r1](https://reviewable.io/reviews/harness/harness-core/40871#-NI8LsRz9axoJcKauSh7:-NIl4gGK1UUhIko_au6x:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/be81442f72ed6883af9c0bee2b62d664409ecc72/420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java#L21)):*

why helper. Lets just call it cache.
</blockquote></details>

Done.
___
*[`420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java` line 44 at r1](https://reviewable.io/reviews/harness/harness-core/40871#-NI8LgNI3R-7BJa9hcAg:-NIl4fxQ67EpfwrTZdZc:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/be81442f72ed6883af9c0bee2b62d664409ecc72/420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java#L44)):*

I think this method should be in metrics publisher to avoid bringing in metricsService as a dependency. 
</blockquote></details>

Done.
___
*[`420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java` line 24 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NI8nnES3BpEtdzuDqDz:-NIl4h1QCSpKrGSC8rRU:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java#L24)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

> r create a new service for delegateTaskCacheService.
this is a good suggestion, we have lot many caches and to use all of them wisely, make sense to make a one unified cache. May be we can take that up as a separate project @jennyjames21 

</blockquote></details>

Done.
___
*[`420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java` line 49 at r2](https://reviewable.io/reviews/harness/harness-core/40871#-NIGU_qb4IK9oPHAOYOP:-NIl4i452yFmTJBVvUSa:bq8xdav) ([raw file](https://github.com/harness/harness-core/blob/3165aa8d8b0da53c391f1f9974e7774b8d2349aa/420-delegate-service/src/main/java/io/harness/delegate/utils/DelegateTasksCountCacheHelper.java#L49)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

will iterating over this map, incurs some cache penalty which can increase cache lookup time ?
</blockquote></details>

No, this creates a separate map, and then it will iterate over the map.

gaurav-jain1 (dismissed):
> Reviewed 1 of 4 files at r2, 3 of 10 files at r3, 10 of 10 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: 1 change requests. 5. unresolved discussions.

gaurav-jain1 (approved):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: 1 change requests. 5. unresolved discussions.

mdmilic (dismissed):
> Reviewed 3 of 10 files at r3, 7 of 10 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: 2. unresolved discussions. 

___
*[`420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java` line 145 at r6](https://reviewable.io/reviews/harness/harness-core/40871#-NIqj4PX6f-4cn280CNl:-NIqj4PX6f-4cn280CNm:b-mrvde3) ([raw file](https://github.com/harness/harness-core/blob/8d7fdc5d0b779d9249c50cb11e4efdeaee6581ae/420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java#L145)):*
> ```Java
>   private void recordDelegateTasksByRankMetrics() {
>     try {
>       for (Map.Entry<String, Long> entry : delegateCache.getMapOfOptionalDelegateTasksCount().entrySet()) {
> ```

Don't want to block it just for this, but if you have to make some other changes, change the signature of this method:
* Don't use data structure names
* No need to contain word delegate
* Have single method which takes rank as a parameter instead of two methods.
It could be called something like `getTaskCountPerAccount(Rank rank)`



_Code quote:_
```Java
getMapOfOptionalDelegateTasksCount
```
___
*[`420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java` line 162 at r6](https://reviewable.io/reviews/harness/harness-core/40871#-NIqh4k62znknaRLc0A-:-NIqh4k62znknaRLc0A0:b8rz395) ([raw file](https://github.com/harness/harness-core/blob/8d7fdc5d0b779d9249c50cb11e4efdeaee6581ae/420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java#L162)):*
> ```Java
> 
>   private void recordDelegateTaskMetrics(String accountId, String taskRank, long value) {
>     try (DelegateAccountMetricContext ignore = new DelegateAccountMetricContext(accountId);
> ```

I haven't seen it being used like this before for metrics context. Possible it would work but double check. Usually we would create a single metric context with both rank & accountId



_Code quote:_
```Java
    try (DelegateAccountMetricContext ignore = new DelegateAccountMetricContext(accountId);
         DelegateTaskRankMetricContext ignore1 = new DelegateTaskRankMetricContext(taskRank)) {
```
___
*[`420-delegate-service/src/main/java/io/harness/service/impl/DelegateCacheImpl.java` line 229 at r6](https://reviewable.io/reviews/harness/harness-core/40871#-NIqkOmk-UKyCSSVVAtr:-NIqkOmk-UKyCSSVVAts:bskzzle) ([raw file](https://github.com/harness/harness-core/blob/8d7fdc5d0b779d9249c50cb11e4efdeaee6581ae/420-delegate-service/src/main/java/io/harness/service/impl/DelegateCacheImpl.java#L229)):*
> ```Java
> 
>   @Override
>   public long getOptionalDelegateTasksCount(@NotNull String accountId) {
> ```

Same, no need for multiple methods, one which takes rank as arg is better. Name like `getTaskCount` is more than enough to know what it does.

arpit3107 (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: 5 file(s) to review. 1. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java` line 145 at r6](https://reviewable.io/reviews/harness/harness-core/40871#-NIqj4PX6f-4cn280CNl:-NJ3x_169s1U67FIadCD:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/8d7fdc5d0b779d9249c50cb11e4efdeaee6581ae/420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java#L145)):*

Don't want to block it just for this, but if you have to make some other changes, change the signature of this method:
* Don't use data structure names
* No need to contain word delegate
* Have single method which takes rank as a parameter instead of two methods.
It could be called something like `getTaskCountPerAccount(Rank rank)`
</blockquote></details>

Done
___
*[`420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java` line 162 at r6](https://reviewable.io/reviews/harness/harness-core/40871#-NIqh4k62znknaRLc0A-:-NJ3xaNn31ggNznj5J24:bv9tx60) ([raw file](https://github.com/harness/harness-core/blob/8d7fdc5d0b779d9249c50cb11e4efdeaee6581ae/420-delegate-service/src/main/java/io/harness/metrics/impl/DelegateMetricsPublisher.java#L162)):*

I haven't seen it being used like this before for metrics context. Possible it would work but double check. Usually we would create a single metric context with both rank & accountId
</blockquote></details>

Yes this works. I have tested on local
___
*[`420-delegate-service/src/main/java/io/harness/service/impl/DelegateCacheImpl.java` line 229 at r6](https://reviewable.io/reviews/harness/harness-core/40871#-NIqkOmk-UKyCSSVVAtr:-NJ3xeQeEi_6sXfrzhWt:bb7f2mf) ([raw file](https://github.com/harness/harness-core/blob/8d7fdc5d0b779d9249c50cb11e4efdeaee6581ae/420-delegate-service/src/main/java/io/harness/service/impl/DelegateCacheImpl.java#L229)):*

Same, no need for multiple methods, one which takes rank as arg is better. Name like `getTaskCount` is more than enough to know what it does.
</blockquote></details>

done

gaurav-jain1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: 5 file(s) to review. 1 change requests. 1. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 231 at r7](https://reviewable.io/reviews/harness/harness-core/40871#-NJ4IOD94_ING1BCspsg:-NJ4IODA8l2NogDyu5DC:b-utlb59) ([raw file](https://github.com/harness/harness-core/blob/5506bdae28a7379761b9098abbc7afd762c77c98/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L231)):*
> ```Java
>   public static final long MAX_OPTIONAL_DELEGATE_TASKS_COUNT_PER_ACCOUNT = 5000;
> 
>   public static final long MAX_IMPORTANT_DELEGATE_TASKS_COUNT_PER_ACCOUNT = 5000;
> ```

We can consider taking these limits from config map.



_Code quote:_
```Java
  public static final long MAX_OPTIONAL_DELEGATE_TASKS_COUNT_PER_ACCOUNT = 5000;

  public static final long MAX_IMPORTANT_DELEGATE_TASKS_COUNT_PER_ACCOUNT = 5000;
```

mdmilic (dismissed):
> Reviewed 3 of 5 files at r7, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: :shipit: complete! ✓

gaurav-jain1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: 1 file(s) to review. 1 change requests. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 354 at r9](https://reviewable.io/reviews/harness/harness-core/40871#-NJ95oxhAL1RsYSn0tOu:-NJ95oxhAL1RsYSn0tOv:b1ps4yg) ([raw file](https://github.com/harness/harness-core/blob/baf42d6190dbacb147f188089ba825c5d1afc1ea/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L354)):*
> ```Java
>       case IMPORTANT:
>         currentTaskCount = delegateCache.getTasksCount(task.getAccountId(), DelegateTaskRank.IMPORTANT);
>         maxTaskCount = mainConfiguration.getPortal().getImportantDelegateTaskRejectAtLimit();
> ```

What if we didn't set this value and turns out to be 0 or -1. Assume some static value,
mainConfiguration.getPortal().getImportantDelegateTaskRejectAtLimit()  >=0 ? mainConfiguration.getPortal().getImportantDelegateTaskRejectAtLimit() : MAX_IMPORTANT_DELEGATE_TASKS_COUNT_PER_ACCOUNT

arpit3107 (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: 1 file(s) to review. 1 change requests. 2. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 231 at r7](https://reviewable.io/reviews/harness/harness-core/40871#-NJ4IOD94_ING1BCspsg:-NJ9B_YrAAug8cnfJl-N:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/5506bdae28a7379761b9098abbc7afd762c77c98/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L231)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

We can consider taking these limits from config map.
</blockquote></details>

Done.
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 354 at r9](https://reviewable.io/reviews/harness/harness-core/40871#-NJ95oxhAL1RsYSn0tOu:-NJ9Ba664wlB_x_4d2yX:barn2w6) ([raw file](https://github.com/harness/harness-core/blob/baf42d6190dbacb147f188089ba825c5d1afc1ea/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L354)):*
<details><summary><i>Previously, gaurav-jain1 (Gaurav Jain) wrote…</i></summary><blockquote>

What if we didn't set this value and turns out to be 0 or -1. Assume some static value,
mainConfiguration.getPortal().getImportantDelegateTaskRejectAtLimit()  >=0 ? mainConfiguration.getPortal().getImportantDelegateTaskRejectAtLimit() : MAX_IMPORTANT_DELEGATE_TASKS_COUNT_PER_ACCOUNT
</blockquote></details>

the value by default is present in config.yml of cg-manager. And we should not safeguard if by mistake we mark this value as -1. This limit was there earlier also, so i can assume that its correctly set at all the places.

gaurav-jain1 (approved):
> Reviewed 3 of 5 files at r7, 1 of 2 files at r8, 1 of 1 files at r9, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: :shipit: complete! ✓

AnupamIO (approved):
> Reviewed 1 of 4 files at r2, 2 of 10 files at r3, 5 of 10 files at r5, 3 of 5 files at r7, 1 of 2 files at r11, 1 of 1 files at r12, 3 of 7 files at r13, 4 of 4 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/40871)* status: :shipit: complete! ✓

PR #41246, opened 2022-12-07
fix:[CDS-47909] Silent ignore of model mapping exception

Fernando-Dourado:
## Describe your changes
A few months ago we saw issues during the conversion of object types, more information at [CDS-40452: CG - It takes a long time to abort an execution](https://harness.atlassian.net/browse/CDS-40452). At that time we choose to ignore the exception and proceed with the abort operation, now my plan is to do the same.

The current exception doesn’t show which field has a wrong value and breaks the workflow execution. We are going to ignore the exception and add a piece of better logging information to track the root cause if it happens again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/41246)


abhinavmittal-harness (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41246)* status: 3 file(s) to review. 1 of 2 approvals obtained.

Y0709 (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41246)* status: 1 file(s) to review. 1 of 2 approvals obtained.

400-rest/src/main/java/software/wings/sm/StateMachineExecutor.java
```diff
@@ -717,7 +717,13 @@ public State getStateForExecution(ExecutionContextImpl context, StateExecutionIn
     log.info("startStateExecution for State {} of type {}", currentState.getName(), currentState.getStateType());
 
     if (stateExecutionInstance.getStateParams() != null) {
-      MapperUtils.mapObject(stateExecutionInstance.getStateParams(), currentState);
+      try {
+        MapperUtils.mapObject(stateExecutionInstance.getStateParams(), currentState);
+      } catch (org.modelmapper.MappingException e) {
+        log.error(String.format("Got model mapping exception during mapping the stateParams %s [currentState=%s]",
+                      stateExecutionInstance.getStateParams(), currentState),
+            e);
```
lasdox, line 725: why are you not rethrowing exception here?

Fernando-Dourado: To avoid execution fail, the same approach is made at https://harness.atlassian.net/browse/CDS-40452.

PR #41773, opened 2022-12-20
feat: [PL-30496]: Implement Redis Lock based Batch mode for Iterators framework

Raghav-Murali:
## Describe your changes

The existing Pump and Loop modes in the iterator frameworks carryout `findAndModify` operation from all the pods in which they are running. This causes a lot of write conflicts and increases the load on the Mongo cluster. Another major disadvantage with `findAndModify` is that it operates on one document at a time thus it can cause significant delays in processing the documents and network overhead. 

To overcome the above disadvantage of the existing iterator framework a new Redis Lock based mode has been implemented which will ensure that multiple pods are not working on the same set of documents. In this mode we will also work on batches of documents thereby optimizing the iterator delay and network overheads. This is achieved by running a `find` query to first get a batch of documents to work on and then carrying out a bulkWrite operation to update the `nextIteration` field of the fetched documents.

The Redis Lock based mode design has been captured in this page - https://harness.atlassian.net/wiki/spaces/PLATFORM/pages/21247590403/Mongo+Persistence+Iterators+-+New+modes

This PR has been tested in the Pre-QA Env covering the below test scenarios -
1. Redis Batch Mode iterator starting.
2. Fetching of batch docs from Mongo and Bulk writing batch docs to Mongo.
3. All the docs / records in ArtifactStream collection have been processed or not by checking the value of the `nextIteration` field.  

## Checklist

## Comment Triggers
<details>

</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
- SonarScan: `trigger ss`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)


Raghav-Murali:
trigger feature-build

raghuAtWings:
*[`400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java` line 104 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvIGf72KcN8bAJwSJ:-NJuvIGf72KcN8bAJwSK:b-r9rf8w) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java#L104)):*
> ```Java
>                     .persistenceProvider(persistenceProvider)
>                     .filterExpander(
>                         query -> query.field(ArtifactStreamKeys.collectionEnabled).in(Arrays.asList(true, null))));
> ```

why null in asList?

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 28 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvQ_RAEiblw9DvOhC:-NJuvQ_RAEiblw9DvOhD:b-b8qjs5) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L28)):*
> ```Java
>   private final HashMap<String, IteratorBaseHandler> iteratorHandlerMap;
>   @Getter private final HashMap<String, IteratorState> iteratorState;
>   PersistentLocker persistentLocker;
> ```

make it private

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 176 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvZR29EX675AFDE4Y:-NJuvZR29EX675AFDE4Z:b-v04pia) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L176)):*
> ```Java
>     if (configOption.isEnabled()) {
>       log.info("Iterator {} is enabled - starting it up", configOption.getName());
>       if (configOption.getIteratorMode().equals("REDIS_BATCH")) {
> ```

consider using switch here

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/iterator/PersistenceIteratorFactory.java` line 139 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvllnCyGIm694hwJK:-NJuvllnCyGIm694hwJL:b-17bkjc) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/iterator/PersistenceIteratorFactory.java#L139)):*
> ```Java
>   createRedisBatchIteratorWithDedicatedThreadPool(
>       RedisBatchExecutorOptions options, Class<?> cls, MongoPersistenceIteratorBuilder<T, F> builder) {
>     if (!workersConfiguration.confirmWorkerIsActive(cls)) {
> ```

what does this mean?

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 67 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvrVM4sVnc4YkK33Q:-NJuvrVM4sVnc4YkK33R:bg175my) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L67)):*
> ```Java
>   private static final Duration QUERY_TIME = ofMillis(200);
>   private static final String SEMAPHORE_ACQUIRE_ERROR = "Working on entity was interrupted";
>   private static final int LOCK_TIMEOUT = 5; // The lockTimeout is the duration a lock is held
> ```

the name should reflect the unit (e.g. LOCK_TIMEOUT_SECONDS)

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 363 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJvcygBAlTjyn9-a-yT:-NJvcygBAlTjyn9-a-yU:bg0sywn) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L363)):*
> ```Java
>    * Process method for Redis Batch mode iterator
>    */
>   public void redisBatchProcess() {
> ```

couldn't find where we are processing the docs based on the serial number of the iterator pod.

raghuAtWings:
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 363 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJvcygBAlTjyn9-a-yT:-NJvcygBAlTjyn9-a-yU:bg0sywn) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L363)):*
> ```Java
>    * Process method for Redis Batch mode iterator
>    */
>   public void redisBatchProcess() {
> ```

couldn't find where we are processing the docs based on the serial number of the iterator pod.

Raghav-Murali:
trigger messagecheck

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: 16 file(s) to review. 6. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java` line 104 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvIGf72KcN8bAJwSJ:-NJw6iV_6IfVlxDSV9Gu:b-j3z2w5) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/400-rest/src/main/java/io/harness/workers/background/critical/iterator/ArtifactCollectionHandler.java#L104)):*

why null in asList?
</blockquote></details>

This is to ensure we pick the documents that do not have a `collectionEnabled` field or has `collectionEnabled` as true.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 28 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvQ_RAEiblw9DvOhC:-NJw6x_F0b2RoDkoOpg9:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L28)):*

make it private
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 176 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvZR29EX675AFDE4Y:-NJwA7Lp6oklJJPww2pa:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L176)):*

consider using switch here
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/PersistenceIteratorFactory.java` line 139 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvllnCyGIm694hwJK:-NJwCsqV68K_WZWeuzhn:b-ruvyr2) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/iterator/PersistenceIteratorFactory.java#L139)):*

what does this mean?
</blockquote></details>

This is to check if the Handler for this iterator is active or not. Whenever a document is retrieved from Mongo it is submitted to the Handler to process it. And if the Handler is not active we will not proceed further as the doc cannot be processed.

By default the Handler is active. However, if the package to which the Handler belongs to is made inactive then all Handlers in that package will be inactive as well.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 67 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvrVM4sVnc4YkK33Q:-NJwBT986sA1Uhl7-Ljc:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L67)):*

the name should reflect the unit (e.g. LOCK_TIMEOUT_SECONDS) 
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 363 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJvcygBAlTjyn9-a-yT:-NJwBl_4Cbq0YcIHc9oa:b-9d1l8p) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L363)):*

couldn't find where we are processing the docs based on the serial number of the iterator pod. 
</blockquote></details>

In this method we will not deploy as `K8s kind: statefulSet` which will give serial number to the pod. Instead we will deploy as `K8s kind: deployment` as before and use a distributed lock in Redis Memstore to synchronize across pods on who can access the next batch of documents. 

All the pods will try to acquire a Redis Lock and whoever gets the lock will get next batch of documents (ordered by the iteration field), update its iteration field to a future time (given by targetInterval) and release the lock. The next pod will now get the next batch of docs and so on. Now since only one pod is writing to Mongo it will not cause writeConflicts and since we're working on batches it should be faster.

gn-harness (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: Some of the checks are too old. 16 file(s) to review. 1 change requests. 6. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 48 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKOz_6N8p6sLVxCBHCa:-NKOz_6N8p6sLVxCBHCb:b49iv5e) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L48)):*
> ```Java
>   protected abstract void createAndStartIterator(
>       PersistenceIteratorFactory.PumpExecutorOptions executorOptions, Duration targetInterval);
> ```

Please add a java doc to this one
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 28 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvQ_RAEiblw9DvOhC:-NKP-BKF-FuZMQ4kfuQk:b4kau9o) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L28)):*

Done
</blockquote></details>

and final too
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 184 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP6OoDEZ-577UauKbv:-NKP6OoDEZ-577UauKbw:b-6cli2c) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L184)):*
> ```Java
>                       .poolSize(configOption.getThreadPoolSize())
>                       .interval(getIntervalDuration(configOption.getTargetIntervalInSeconds()))
>                       .persistentLocker(persistentLocker)
> ```

Why is this persistentLocker being passed around and now created in the MongoPersistenceIterator



_Code quote:_
```Java
.persistentLocker(persistentLocker)
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 315 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP78jvDbCEB-QfcXi6:-NKP78jvDbCEB-QfcXi7:b1mtmzy) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L315)):*
> ```Java
>    * @return AcquiredLock
>    */
>   private AcquiredLock acquireLock() {
> ```

keep private methods after public



_Code quote:_
```Java
 private A
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 334 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP81zG52zNfoNVaFlk:-NKP81zG52zNfoNVaFll:btq1f71) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L334)):*
> ```Java
>    * @param docIds List of document Ids
>    */
>   public void updateDocumentNextIteration(List<String> docIds) {
> ```

why is this public?



_Code quote:_
```Java
public void updateDocumentNextIteration
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 362 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP8AbXDrwoMOYYX1Xu:-NKP8AbXDrwoMOYYX1Xv:bij0ajb) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L362)):*
> ```Java
> 
>   /**
>    * Process method for Redis Batch mode iterator
> ```

add some details on what this method does



_Code quote:_
```Java
Process method for Redis Batch mode iterator
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 368 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP8KJa5OIq0VQ4R5X6:-NKP8KJa5OIq0VQ4R5X7:b2rx9ex) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L368)):*
> ```Java
>     // Set the batchSize to fetch documents as 2x number of worker threads.
>     // The main thread is fetching documents so worker thread count is N-1.
>     int batchSize = Math.max(BATCH_SIZE_INCREASE * (threadPoolExecutor.getCorePoolSize() - 1), 1);
> ```

this variable name is confusing. This is not even config driven yet nor is this variable reused, might as well replace with just a number 2



_Code quote:_
```Java
ATCH_SIZE_INCREASE
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 380 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP9qn50dYmHD9uhUn1:-NKP9qn50dYmHD9uhUn2:bbt476s) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L380)):*
> ```Java
>       int limit = Math.min(batchSize, batchSize - threadPoolExecutor.getQueue().size());
> 
>       if (limit <= 0) {
> ```

nit: If very little has been processed till now, we will try to fetch very few documents in the batch. Watch out for this and probably add a min_bound for making a query - unless not hit, sleep



_Code quote:_
```Java
  if (limit <= 0) {
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 396 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPD76j0EJVSd2MtFfu:-NKPD76j0EJVSd2MtFfv:b-2to5n8) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L396)):*
> ```Java
>         Thread.currentThread().interrupt();
>         return;
>       }
> ```

Why is this required? Which resource we are attempting to lock here?



_Code quote:_
```Java
        semaphore.acquire();
      } catch (InterruptedException e) {
        log.error(SEMAPHORE_ACQUIRE_ERROR, e);
        iteratorMetricsService.recordIteratorMetrics(iteratorName, ITERATOR_ERROR);
        Thread.currentThread().interrupt();
        return;
      }
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 402 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPDhVs2AAVLzubUQyf:-NKPDhVs2AAVLzubUQyg:b-30cvnv) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L402)):*
> ```Java
> 
>       // Acquire the distributed lock
>       AcquiredLock acquiredLock = acquireLock();
> ```

This is not inside the try / with us releasing lock in finally, might want to move it inside?



_Code quote:_
```Java
 AcquiredLock acquiredLock = acquireLock();
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 427 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPCXyj99QtXvnv1GMl:-NKPCXykCpxp7mI1jp7u:b-tz09ws) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L427)):*
> ```Java
>         } else {
>           // Update the documents next iteration field
>           updateDocumentNextIteration(docIds);
> ```

nit:
updateDocumentNextIteration(docIds);
docsAvailable = !docIds.isEmpty()


def updateDocumentNextIteration():
  if docIds.isEmpty() {
     return;
   }
   ...



_Code quote:_
```Java
        if (docIds.isEmpty()) {
          // There are no more documents available so break from loop
          docsAvailable = false;
        } else {
          // Update the documents next iteration field
          updateDocumentNextIteration(docIds);
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 470 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPEUpSD7QRIHqeb56G:-NKPEUpSD7QRIHqeb56H:bjs8nhe) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L470)):*
> ```Java
>    * @param entity - Mongo document that worker thread should process
>    */
>   public void processOneEntity(T entity) {
> ```

please double check visibility of all methods



_Code quote:_
```Java
public
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 481 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPF9HLCF-NrgkNASUk:-NKPF9HLCF-NrgkNASUl:b-ekowt) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L481)):*
> ```Java
>       }
> 
>       long startTime = currentTimeMillis();
> ```

This is different from processEntity?



_Code quote:_
```Java
    try (EntityLogContext ignore = new EntityLogContext(entity, OVERRIDE_ERROR)) {
      try {
        semaphore.acquire();
      } catch (InterruptedException e) {
        log.error(SEMAPHORE_ACQUIRE_ERROR, e);
        iteratorMetricsService.recordIteratorMetrics(iteratorName, ITERATOR_ERROR);
        Thread.currentThread().interrupt();
        return;
      }

      long startTime = currentTimeMillis();
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceProvider.java` line 122 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPG3-YE1sMLgZWz6jF:-NKPG3-YE1sMLgZWz6jG:bcewebt) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceProvider.java#L122)):*
> ```Java
> 
>   @Override
>   public BulkWriteResult bulkWriteDocumentsMatchingIds(
> ```

possible to share code with MongoPersistenceRequiredProvider?



_Code quote:_
```Java
  @Override
  public BulkWriteResult bulkWriteDocumentsMatchingIds(
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceRequiredProvider.java` line 119 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPFikM5kEEmvM64EBX:-NKPFikN8a1SqKFEDowH:b8wlve6) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceRequiredProvider.java#L119)):*
> ```Java
> 
>   @Override
>   public BulkWriteResult bulkWriteDocumentsMatchingIds(
> ```

possible to write some tests?



_Code quote:_
```Java
 public BulkWriteResult bulkWriteDocumentsMatchingIds(
```

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: Some of the checks are too old. 16 file(s) to review. 1 change requests. 20. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 48 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKOz_6N8p6sLVxCBHCa:-NKQJVed7LieIfU8_dvF:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L48)):*

Please add a java doc to this one
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 28 at r3](https://reviewable.io/reviews/harness/harness-core/41773#-NJuvQ_RAEiblw9DvOhC:-NKQJX9v3nUe8dHZ8Pkq:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/9c27466e3c71dacb5644b6c32cc1dd6adad46d8f/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L28)):*

and final too
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 184 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP6OoDEZ-577UauKbv:-NKQKFdy13rr21_qG9gr:b-i2hbqk) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L184)):*

Why is this persistentLocker being passed around and now created in the MongoPersistenceIterator
</blockquote></details>

The persistentLocker is created in `WingsApplication` and pass it to `IteratorExecutionHandler` which is starting all the iterators. The `IteratorExecutionHandler` will pass it to the `MongoPersistenceIterator` which is the final location where this lock is required. Since the iterators are now being maintained by the `IteratorExecutionHandler` which adds an extra hop in between to pass this persistentLocker.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 315 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP78jvDbCEB-QfcXi6:-NKQLRf46ZCTunm2ldBB:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L315)):*

keep private methods after public
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 334 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP81zG52zNfoNVaFlk:-NKQM3aN4Rmgh8PCSfQ2:b1lufkf) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L334)):*

why is this public?
</blockquote></details>

Actually this is not needed elsewhere. I will make it private.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 362 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP8AbXDrwoMOYYX1Xu:-NKQMDFv8O2wTH1UUtOU:bx5jxrt) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L362)):*

add some details on what this method does
</blockquote></details>

Will add a description.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 368 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP8KJa5OIq0VQ4R5X6:-NKQPgk90BsNSKn96wBK:b-1s7iva) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L368)):*

this variable name is confusing. This is not even config driven yet nor is this variable reused, might as well replace with just a number 2
</blockquote></details>

I got a SonarQube code analysis warning to not use magic numbers that is why I updated it to a Constant. Shall I rename it to BATCH_SIZE_MULTIPLY or BATCH_SIZE_AUGMENT ?
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 380 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP9qn50dYmHD9uhUn1:-NKQRPg4-ohLMLfT4v-L:b-mb99c2) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L380)):*

nit: If very little has been processed till now, we will try to fetch very few documents in the batch. Watch out for this and probably add a min_bound for making a query - unless not hit, sleep
</blockquote></details>

Ah, yes. Thanks for highlighting this. Will add a min bound to ensure we do not end up fetching very few documents. I will make it a constant for now.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 396 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPD76j0EJVSd2MtFfu:-NKQQnNKAAWhZ4Oqk7ci:bgg376m) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L396)):*

Why is this required? Which resource we are attempting to lock here?
</blockquote></details>

Currently, `redisBatchProcess` is invoked using `scheduleAtFixedRate` function of the executor so only one thread will be executing this method at a time. However, if someone invokes `redisBatchProcess` through schedule or submit methods of the executor then multiple threads can execute this method simultaneously. So this is just a safety check to ensure that not more than the configured number of threads run simultaneously.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 402 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPDhVs2AAVLzubUQyf:-NKQRuhn-N0FbhgYF3oT:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L402)):*

This is not inside the try / with us releasing lock in finally, might want to move it inside?
</blockquote></details>

Done
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 427 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPCXyj99QtXvnv1GMl:-NKQSubl6QGvR1vRIf7L:b-r8mx0y) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L427)):*

nit:
updateDocumentNextIteration(docIds);
docsAvailable = !docIds.isEmpty()


def updateDocumentNextIteration():
  if docIds.isEmpty() {
     return;
   }
   ...
</blockquote></details>

Ah, this is neat. Done.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 470 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPEUpSD7QRIHqeb56G:-NKQT0Bt7Vf4IacOyn7i:bhmry0w) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L470)):*

please double check visibility of all methods
</blockquote></details>

Yeah, this need not be public. I will update it for the new methods added.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 481 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPF9HLCF-NrgkNASUk:-NKQTI7j0ff8knAfygvo:b-jqeqdm) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L481)):*

This is different from processEntity?
</blockquote></details>

Yes. The main functionality I wanted was not to wait until a worker picks up the task from the Queue and thus don't need the wait and notify semantics. In processEntity we wait for a worker to pick up the doc from Q whereas in this we don't wait for it and thus we don't use wait / notify. Should I give this function a better name instead ?
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceProvider.java` line 122 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPG3-YE1sMLgZWz6jF:-NKQU2yc7qPehxjBJvFy:baxdpo4) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceProvider.java#L122)):*

possible to share code with MongoPersistenceRequiredProvider?
</blockquote></details>

Yeah, I was wondering how I can share the same code between MongoPersistenceProvider and MongoPersistenceRequiredProvider. Currently I just duplicated it as it was easier. Shall I add a new helper class that has all the new Mongo implementations and invoke that from both MongoPersistenceProvider and MongoPersistenceRequiredProvider ? 
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceRequiredProvider.java` line 119 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPFikM5kEEmvM64EBX:-NKQUWbJ8jAJ3n3uVnBJ:b-27q8iy) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceRequiredProvider.java#L119)):*

possible to write some tests?
</blockquote></details>

I should be able to mock the Mongo part and add few tests. Let me look into it a bit more and will update.

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: Some of the checks are too old. 16 file(s) to review. 1 change requests. 20. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceProvider.java` line 122 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPG3-YE1sMLgZWz6jF:-NKV2i828kkZuBBy5Pn5:b-b08831) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceProvider.java#L122)):*

Yeah, I was wondering how I can share the same code between MongoPersistenceProvider and MongoPersistenceRequiredProvider. Currently I just duplicated it as it was easier. Shall I add a new helper class that has all the new Mongo implementations and invoke that from both MongoPersistenceProvider and MongoPersistenceRequiredProvider ? 
</blockquote></details>

Looks like the MongoPersistenceProvider and MongoPersistenceRequiredProvider are exactly same except for the createQuery() function. In MongoPersistenceProvider the createQuery() will include an addition `doesNotExists` clause which the MongoPersistenceRequiredProvider does not include. Apart from this difference both the classes are just replicating the same code. I think maybe I can raise a separate ticket to clean this up later.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceRequiredProvider.java` line 119 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPFikM5kEEmvM64EBX:-NKV0pEwC-2dzY-qk_gb:bax79vg) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceRequiredProvider.java#L119)):*

I should be able to mock the Mongo part and add few tests. Let me look into it a bit more and will update. 
</blockquote></details>

I looked into it more. Even if I add mocks for the Mongo parts there won't be much that can be tested since the BulkWriteResult that is being returned from this function also has to be mocked. So effectively we won't be testing much. All the new functions added in this class just use a lot of Mongo APIs.

raghuAtWings (approved):
> Reviewed 6 of 16 files at r1, 5 of 6 files at r3, 2 of 2 files at r4, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: 16 file(s) to review. 1 change requests. 14. unresolved discussions. 1 of 2 approvals obtained.

gn-harness (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: Some of the checks are too old. 16 file(s) to review. 1 change requests. 1. unresolved discussions. 1 of 2 approvals obtained. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 60 at r6](https://reviewable.io/reviews/harness/harness-core/41773#-NKtwgZ2ERCWr-K76o9A:-NKtwgZ2ERCWr-K76o9B:b-y0jz2i) ([raw file](https://github.com/harness/harness-core/blob/f78ee42eb310924603c1e93eade62f34ad945200/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L60)):*
> ```Java
>   protected void createAndStartRedisBatchIterator(
>       PersistenceIteratorFactory.RedisBatchExecutorOptions executorOptions, Duration targetInterval) {
>     log.warn("createAndStartShardIterator should be overridden by the child class");
> ```

mark it abstract?
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 184 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP6OoDEZ-577UauKbv:-NKtx4em19qEZRaWEdrL:bq1qy1z) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L184)):*

The persistentLocker is created in `WingsApplication` and pass it to `IteratorExecutionHandler` which is starting all the iterators. The `IteratorExecutionHandler` will pass it to the `MongoPersistenceIterator` which is the final location where this lock is required. Since the iterators are now being maintained by the `IteratorExecutionHandler` which adds an extra hop in between to pass this persistentLocker.
</blockquote></details>

Why not create a new instance of the persistent locker per iterator? or probably just create a singleton instance of PersistentLocker

Actually, if you see the PersistentLocker module (https://github.com/harness/harness-core/blob/develop/960-persistence/src/main/java/io/harness/lock/PersistentLockModule.java#L74)  -- it is already marked to be singleton. So, if you try doing something like 

PersistentLocker persistentLocker = injector.getInstance(Key.get(PersistentLocker.class));

inside the iterator, we should get the same instance.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 380 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP9qn50dYmHD9uhUn1:-NKtzVYgBWuRFEEM_VH0:b-lf9pm) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L380)):*

Ah, yes. Thanks for highlighting this. Will add a min bound to ensure we do not end up fetching very few documents. I will make it a constant for now.
</blockquote></details>

nit: you should probably wait until min batch size can be handled.



_Code quote:_
```Java
REDIS_MIN_BATCH_SIZE)
```
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 396 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPD76j0EJVSd2MtFfu:-NKtzsIR4xTMpp3JIdYz:b-1ey6l4) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L396)):*
> attempting
Can you please document it here?

Also the semaphore acquired in the processOneBatch same?
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 481 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPF9HLCF-NrgkNASUk:-NKu09qA5DlsMtEbmA0j:bgn9ijv) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L481)):*

Yes. The main functionality I wanted was not to wait until a worker picks up the task from the Queue and thus don't need the wait and notify semantics. In processEntity we wait for a worker to pick up the doc from Q whereas in this we don't wait for it and thus we don't use wait / notify. Should I give this function a better name instead ?
</blockquote></details>

There is a lot of code duplication here, can we reduce the duplication somehow?

Any downsides of not using wait / notify? Also, if this is the main difference, then name the function accordingly

and can you please document (and refresh my memory) why we dont need to wait here? 
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceProvider.java` line 122 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPG3-YE1sMLgZWz6jF:-NKu0_ty7wRfBgAfUBUH:b-8xehil) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceProvider.java#L122)):*

Looks like the MongoPersistenceProvider and MongoPersistenceRequiredProvider are exactly same except for the createQuery() function. In MongoPersistenceProvider the createQuery() will include an addition `doesNotExists` clause which the MongoPersistenceRequiredProvider does not include. Apart from this difference both the classes are just replicating the same code. I think maybe I can raise a separate ticket to clean this up later.
</blockquote></details>

separate sounds ok to me - just create a jira to track
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceRequiredProvider.java` line 119 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPFikM5kEEmvM64EBX:-NKu2tc9CuiopVVI1gYB:bg5ypox) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceRequiredProvider.java#L119)):*

I looked into it more. Even if I add mocks for the Mongo parts there won't be much that can be tested since the BulkWriteResult that is being returned from this function also has to be mocked. So effectively we won't be testing much. All the new functions added in this class just use a lot of Mongo APIs.
</blockquote></details>

Ack. We should ideally be able to write tests which spin up local DB and test these interactions - not in this PR though.

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: Some of the checks are too old. 16 file(s) to review. 1 change requests. 7. unresolved discussions. 1 of 2 approvals obtained. 

___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java` line 60 at r6](https://reviewable.io/reviews/harness/harness-core/41773#-NKtwgZ2ERCWr-K76o9A:-NKuYMdF6QH0n0devpxw:bfgmx2c) ([raw file](https://github.com/harness/harness-core/blob/f78ee42eb310924603c1e93eade62f34ad945200/960-persistence/src/main/java/io/harness/iterator/IteratorBaseHandler.java#L60)):*

mark it abstract?
</blockquote></details>

Marking it abstract would require it to be implemented in all the handlers. Since right now we're only adding this support to ArtifactCollection and ArtifactCleanup, I have not marked it as abstract. I will mark it as abstract once the support for this is extended to all Iterators.
___
*[`960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java` line 184 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP6OoDEZ-577UauKbv:-NKuZFQg026um7v-ZTC3:b-8bultq) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/iterator/IteratorExecutionHandlerImpl.java#L184)):*

Why not create a new instance of the persistent locker per iterator? or probably just create a singleton instance of PersistentLocker

Actually, if you see the PersistentLocker module (https://github.com/harness/harness-core/blob/develop/960-persistence/src/main/java/io/harness/lock/PersistentLockModule.java#L74)  -- it is already marked to be singleton. So, if you try doing something like 

PersistentLocker persistentLocker = injector.getInstance(Key.get(PersistentLocker.class));

inside the iterator, we should get the same instance.
</blockquote></details>

Done. Will do it as per our discussion on Zoom.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 380 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKP9qn50dYmHD9uhUn1:-NKyQL1BC4ElkNk9iOMb:bh3m0au) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L380)):*

nit: you should probably wait until min batch size can be handled.
</blockquote></details>

Actually we do wait for few seconds above if the jobQ is full already. If we have come to this part then the jobQ is not full, so I think it should be fine to add REDIS_MIN_BATCH_SIZE (5) docs to the Q.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 396 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPD76j0EJVSd2MtFfu:-NKyR1ih4IpVGzxrYumY:byvn93a) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L396)):*

> attempting
Can you please document it here?

Also the semaphore acquired in the processOneBatch same?
</blockquote></details>

Yes, the semaphore used in `processOneEntity()` is the same. So the main thread is executing `redisBatchProcess()` and it will acquire the semaphore thereby reducing the number of threads in the executor service by 1. So now whatever threads are remaining will be the worker threads which will execute `processOneEntity()`. Each worker will also acquire the semaphore and decrement the count by 1 ensuring only the configured nos of threads are running.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java` line 481 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPF9HLCF-NrgkNASUk:-NKyTd7c-FtkSHTnG4YQ:buqlage) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/MongoPersistenceIterator.java#L481)):*

There is a lot of code duplication here, can we reduce the duplication somehow?

Any downsides of not using wait / notify? Also, if this is the main difference, then name the function accordingly

and can you please document (and refresh my memory) why we dont need to wait here? 
</blockquote></details>

Let me move out all the IteratorMetrics updating code to a common function and reuse them.

I will rename the function to `processEntityWithoutWaitNotify()`. The reason we don't want to wait / notify here is because we want to release the Redis Lock asap and not delay other pods. We are handling Q overflow using a limit logic instead of relying on wait / notify. Have documented this in the java doc for `redisBatchProcess()`.
___
*[`960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceProvider.java` line 122 at r5](https://reviewable.io/reviews/harness/harness-core/41773#-NKPG3-YE1sMLgZWz6jF:-NKyYJ369456sN-DkoIm:b-gysgly) ([raw file](https://github.com/harness/harness-core/blob/c94f2134d22385445386c2a4e2dcebb50c27e970/960-persistence/src/main/java/io/harness/mongo/iterator/provider/MorphiaPersistenceProvider.java#L122)):*

separate sounds ok to me - just create a jira to track
</blockquote></details>

Cool. I have created this ticket to track - https://harness.atlassian.net/browse/PL-30741

Raghav-Murali:
trigger feature-build

Raghav-Murali:
trigger feature-build

Raghav-Murali:
trigger ft1

raghuAtWings (approved):
> Reviewed 2 of 9 files at r7, 5 of 6 files at r8, 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: 15 file(s) to review. 1 change requests. 7. unresolved discussions. 1 of 2 approvals obtained.

gn-harness (approved):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: 15 file(s) to review.

gn-harness (approved):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: 15 file(s) to review.

gn-harness (approved):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: 15 file(s) to review.

abhinav-harness (approved):
> Reviewed 3 of 16 files at r1, 2 of 6 files at r3, 1 of 3 files at r6, 2 of 9 files at r7, 5 of 6 files at r8, 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/41773)* status: :shipit: complete! ✓

PR #42369, opened 2023-01-09
fix:[del-5688]: fixing client tools for immutable delegate

bharatgoelharness:
wiki : https://harness.atlassian.net/wiki/spaces/DEL/pages/21241922781/Delegate+3rd+party+tools+version+upgrade+plan
In this Jira we are optimising the client tools to reduce the cve's.
This was tested by CD team and all the sign off detail is mentioned in this channel **#immutable-delegate-signoff** 

Change is only for immutable delegate and we have uploaded the required new binaries in both QA and PROD cdn.

Created 2 docker files
1: docker-immutable-minimal : This is base minimal file which will have delegate jar and scm
2: docker-immutable: This file will have all the other tools. To bake the image from this file we need to pass an extra argument which is the base image tag.

minimal image build:  docker buildx build --push --platform linux/amd64,linux/arm64 --tag ${REPO}.minimal .
maximal image build: docker buildx build --push --platform linux/amd64,linux/arm64 --tag ${REPO} .

Below is the  docker run command output for both image
Screenshot 2023-01-13 at 11 42 46 PM

## Checklist

## Comment Triggers
<details>

</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
- SonarScan: `trigger ss`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/42369)


bharatgoelharness:
@mdmilic I have closed the previous PR due to conflicts.

Please review this and this change is only for immutable delegates

mdmilic (commented):
@bharatgoelharness lets not diverge the dockerfiles it will make it harder to maintain. Can you please keep it as one Dockerfile and make the client tool installation conditional (if no args provided we build minimal delegate)? 
You can have a separate installation script for immutable delegates if needed. The alternative approach I can think of is having multi stage build (where minimal and regular delegates can be separate stages in single build file, and full delegate is `FROM minimal` and just adds client tools and envs). 
After you get them built, make sure the image size have not changed significantly. 

> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/42369#-:-NLLgK4n8tEIJ1cP-kPv:b-9bb52w)* status: 2 file(s) to review. 0 of 1 approvals obtained.

bharatgoelharness:
@mdmilic We wanted to make it these images public after few days. So that customer who want to build their own image can use this.
thats why we are having 2 separate files, so that if customer can bake image as per their requirement. They will be having full control over it.

Multistage we can try out but having conditional statement will increase the complexity.

bharatgoelharness:
@mdmilic 
I think we can't use multi stage also as we need to images (with tools and without tools). Customer can use anyone of them. 
What I got to know from multi stage doc that it will give you one image tag finally. Other images will be in cache only.

mdmilic:
> We wanted to make it these images public after few days
I still think we should have a single Dockerfile, but this is why I mentioned that minimal delegate is the default option (because most customers customize minimal delegate and they don't need to do anything special to build it). They are still in full control and the build arg is even self explanatory, so anyone building images would not have hard time understanding this. In my opinion having multiple Dockerfiles is both more confusing for users and harder to maintain for us. 

> @mdmilic I think we can't use multi stage also as we need to images (with tools and without tools). Customer can use anyone of them. What I got to know from multi stage doc that it will give you one image tag finally. Other images will be in cache only.

You would just build it twice the same way as today or if having multiple Dockerfiles. This is [example](https://stackoverflow.com/a/60820156) of conditional multistage I'm referring to. 

BTW, I would love if our build flag is called minimal = true|false, but changing flag name is backwards incompatible change and we would have to support both flags.

bharatgoelharness:
> > We wanted to make it these images public after few days
> > I still think we should have a single Dockerfile, but this is why I mentioned that minimal delegate is the default option (because most customers customize minimal delegate and they don't need to do anything special to build it). They are still in full control and the build arg is even self explanatory, so anyone building images would not have hard time understanding this. In my opinion having multiple Dockerfiles is both more confusing for users and harder to maintain for us.
> 
> > @mdmilic I think we can't use multi stage also as we need to images (with tools and without tools). Customer can use anyone of them. What I got to know from multi stage doc that it will give you one image tag finally. Other images will be in cache only.
> 
> You would just build it twice the same way as today or if having multiple Dockerfiles. This is [example](https://stackoverflow.com/a/60820156) of conditional multistage I'm referring to.
> 
> BTW, I would love if our build flag is called minimal = true|false, but changing flag name is backwards incompatible change and we would have to support both flags.

Still this will give you one final image at a time. We would have to build it 2 times.
We have only 1 dependency that is minimal image is base image for image with tools.
Build stage is useful when we have multiple dependency.

bharatgoelharness:
@raghuAtWings @mdmilic Please review this. We wanted to merge it before next branch cut 783xx.

bharatgoelharness:
@mdmilic 
Regarding ARG TARGETARCH, this will take current system arch if not being passed.

Regarding SCM, this is required for CD task.

mdmilic:
> @mdmilic Regarding ARG TARGETARCH, this will take current system arch if not being passed.
I know, but I think it's better to be explicit here. It can lead to some really hard to debug issues otherwise (because it's different behavior on different hosts). If you think this is what we actually want, then I'm fine. 
> Regarding SCM, this is required for CD task.
All tools are needed by some CD tasks :) minimal image should have only executables/deps/libraries that are needed for core delegate operation (i.e. for absolutely every task), so no SCM by default. Also, it didn't have it so far.

bharatgoelharness:
My bad this SCM change came from CI team. If customer is using minimal image for CI task then we didn't had the scm tool for CI task. AFAIK for each CI task SCM is mandatory.

I can add default ARCH and we should add this in doc also that we support linux/amd64 and linux/arm64 as of now.

mdmilic:
> My bad this SCM change came from CI team. If customer is using minimal image for CI task then we didn't had the scm tool for CI task. AFAIK for each CI task SCM is mandatory. https://harness.atlassian.net/browse/CI-6195
> 
> I can add default ARCH and we should add this in doc also that we support linux/amd64 and linux/arm64 as of now.

I assumed it was probably for CI, but still not all our customers are CI, why would we force SCM on CD only customers.

mdmilic:
> > My bad this SCM change came from CI team. If customer is using minimal image for CI task then we didn't had the scm tool for CI task. AFAIK for each CI task SCM is mandatory. https://harness.atlassian.net/browse/CI-6195
> > I can add default ARCH and we should add this in doc also that we support linux/amd64 and linux/arm64 as of now.
> 
> I assumed it was probably for CI, but still not all our customers are CI, why would we force SCM on CD only customers.

You could say tomorrow that every STO task needs snyk or some other tool, so we just bake it in minimal image. That goes against what minimal image is supposed to be. Minimal image is a starting point for you to customize delegate and manage your own tools, so I'm strongly against baking SCM or anything else into it.

bharatgoelharness:
@mdmilic added the default arch also

bharatgoelharness (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/42369)* status: 3 file(s) to review. 1 change requests. 1. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`dockerization/delegate/Dockerfile-minimal` line 29 at r2](https://reviewable.io/reviews/harness/harness-core/42369#-NLuOfe26JwmMWULKt06:-NM1wXLt2EjxXT2Up3eq:bbwr5o3) ([raw file](https://github.com/harness/harness-core/blob/6461e50ef65975a9d0f90ba190cf510b222c1a18/dockerization/delegate/Dockerfile-minimal#L29)):*

Why do we need SCM in the minimal image?
</blockquote></details>

This is the requirement.

bharatgoelharness:
@mdmilic @raghuAtWings Please approve this.

raghuAtWings (approved):
> Reviewed 1 of 3 files at r2, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/42369)* status: 1 change requests. 1. unresolved discussions.

PR #43118, opened 2023-01-26
fix: [DEL-4875]: Fair load of distribution of task

jennyjames21:
## Describe your changes
Summary:
- Support to load balance task and distribute equally between eligible delegates
- Goal here is to make the first delegate broadcast pick more accurate and to reduce  rebroadcast attempts to other eligible delegates
- Make the order of task broadcasting based on the number of the tasks currently running for eligible delegates, the one with the least running tasks will get more priority
- Feature is behind FF, DELEGATE_TASK_LOAD_DISTRIBUTION

Test performed
- Prometheus metric verified equal distribution of tasks between delegates.
- NG Metrics
- CG Metrics
- Task distribution verified with multiple manager nodes and more than one delegate group.

Sanity
- Without FF enabled, end to end k8 task CG and NG task execution flow 

## Checklist

## Comment Triggers
<details>

</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
- SonarScan: `trigger ss`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/43118)


jennyjames21:
CG - Metrics with task load balanced between 4 immutable delegates
image

jennyjames21:
Sanity - k8 end to end flow 
image

raghuAtWings:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 786 at r8](https://reviewable.io/reviews/harness/harness-core/43118#-NN_KQa43S14zBCCnlY-:-NN_KQa43S14zBCCnlY0:bl356dn) ([raw file](https://github.com/harness/harness-core/blob/ede4160bb31036d808525f3a2b34c7ec7dacdcda/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L786)):*
> ```Java
>           getEligibleDelegateListOrderedNumberByTaskAssigned(delegateTask, eligibleListOfDelegates);
>       // if no task currently running then choose random from the list and give priority if group is whitelisted
>       if (isEmpty(eligibleDelegatesSorted)) {
> ```

do we need the isEmpty check?

raghuAtWings:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 824 at r8](https://reviewable.io/reviews/harness/harness-core/43118#-NN_LEYw11C-3XZeN9f3:-NN_LEYw11C-3XZeN9f4:bkwh239) ([raw file](https://github.com/harness/harness-core/blob/ede4160bb31036d808525f3a2b34c7ec7dacdcda/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L824)):*
> ```Java
>       return emptyList();
>     }
>     Map<String, Integer> delegateTaskCountMap = eligibleListOfDelegates.stream().collect(toMap(k -> k, v -> 0));
> ```

rename k and v to reflect what they are

raghuAtWings:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 788 at r8](https://reviewable.io/reviews/harness/harness-core/43118#-NN_MAIZ0igxtnpHk9VU:-NN_MAIZ0igxtnpHk9VV:bx6za7r) ([raw file](https://github.com/harness/harness-core/blob/ede4160bb31036d808525f3a2b34c7ec7dacdcda/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L788)):*
> ```Java
>       if (isEmpty(eligibleDelegatesSorted)) {
>         for (String delegateId : eligibleListOfDelegates) {
>           if (assignDelegateService.isDelegateGroupWhitelisted(delegateTask, delegateId)) {
> ```

should we do this? in that case if there are two different delegate groups, then the first broadcast will heavily be going to the first delegate group that took the task and is whitelisted?

raghuAtWings:
*[`400-rest/src/test/resources/mongo/indexes.txt` line 187 at r8](https://reviewable.io/reviews/harness/harness-core/43118#-NN_MSp68rVxFb_xyVoj:-NN_MSp68rVxFb_xyVok:bi36gl7) ([raw file](https://github.com/harness/harness-core/blob/ede4160bb31036d808525f3a2b34c7ec7dacdcda/400-rest/src/test/resources/mongo/indexes.txt#L187)):*
> ```help
> delegateSequenceConfig {"name": "uniqueDelegateSequenceIdx", "unique": true} {"accountId": 1, "hostName": 1, "sequenceNum": 1}
> delegateTasks {"name": "createdAt_1", "background": true} {"createdAt": 1}
> delegateTasks {"name": "delegateId_1", "background": true} {"delegateId": 1}
> ```

we are filtering on the task status and delegate ids, we should have a compounded index for it. Check with sreehari for optimized index.

raghuAtWings (commented):
> Reviewed 3 of 4 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43118)* status: 0 of 1 approvals obtained.

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 786 at r8](https://reviewable.io/reviews/harness/harness-core/43118#-NN_KQa43S14zBCCnlY-:-NNbTOZCD8ggQdDmkIN2:b-cy39xo) ([raw file](https://github.com/harness/harness-core/blob/ede4160bb31036d808525f3a2b34c7ec7dacdcda/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L786)):*

do we need the isEmpty check?
</blockquote></details>

yes, needed to apply logic to pick up random from if no task for eligible delegates is currently running

jennyjames21:
*[`400-rest/src/test/resources/mongo/indexes.txt` line 187 at r8](https://reviewable.io/reviews/harness/harness-core/43118#-NN_MSp68rVxFb_xyVoj:-NNbVmTY5JkthmQeE9M7:b2nhb2c) ([raw file](https://github.com/harness/harness-core/blob/ede4160bb31036d808525f3a2b34c7ec7dacdcda/400-rest/src/test/resources/mongo/indexes.txt#L187)):*

we are filtering on the task status and delegate ids, we should have a compounded index for it. Check with sreehari for optimized index. 
</blockquote></details>

Ack

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 788 at r8](https://reviewable.io/reviews/harness/harness-core/43118#-NN_MAIZ0igxtnpHk9VU:-NNbW5Us4Wo7yyHWvunN:b-19x5q8) ([raw file](https://github.com/harness/harness-core/blob/ede4160bb31036d808525f3a2b34c7ec7dacdcda/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L788)):*

should we do this? in that case if there are two different delegate groups, then the first broadcast will heavily be going to the first delegate group that took the task and is whitelisted?
</blockquote></details>

we have that currently, it's a short term solution added to fix Informatica imbalanced distribution observed during POC .
We may not need this check anymore but that all depends on how delegates are configured and how many delegates in their account. 

Since this is under FF, maybe we can add this if needed later. 

Or another option is to add this check only if there are more than X (more than 10 or 15)  eligible delegates, that way we can find the middle ground.

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 788 at r8](https://reviewable.io/reviews/harness/harness-core/43118#-NN_MAIZ0igxtnpHk9VU:-NNgudBB1QM-884ofd5u:boe3pfu) ([raw file](https://github.com/harness/harness-core/blob/ede4160bb31036d808525f3a2b34c7ec7dacdcda/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L788)):*

we have that currently, it's a short term solution added to fix Informatica imbalanced distribution observed during POC .
We may not need this check anymore but that all depends on how delegates are configured and how many delegates in their account. 

Since this is under FF, maybe we can add this if needed later. 

Or another option is to add this check only if there are more than X (more than 10 or 15)  eligible delegates, that way we can find the middle ground.

</blockquote></details>

removed whitelist check

raghuAtWings (dismissed):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43118)* status: :shipit: complete! ✓

jennyjames21:
trigger ti1

raghuAtWings (approved):
> Reviewed 2 of 2 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43118)* status: :shipit: complete! ✓

raghuAtWings (dismissed):
> Reviewed 3 of 3 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43118)* status: :shipit: complete! ✓

jennyjames21:
trigger ft1

mdmilic (changes requested):
> Reviewed 1 of 3 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43118)* status: :shipit: complete! ✓

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 794 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNmtmBV9mumGOkb9F0e:-NNmtmBWAc2VrwXrMt05:bizsk89) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L794)):*
> ```Java
>   protected List<String> getEligibleDelegateListOrderedNumberByTaskAssigned(
>       DelegateTask delegateTask, List<String> eligibleListOfDelegates) {
>     List<DelegateTask> delegateTasks = persistence.createQuery(DelegateTask.class)
> ```

Is this done for every task assignment or just e.g. sync tasks (do only sync tasks have first attempt unicast)? I think it might have big impact on mongo. We can try it out, but we should keep an eye on mongo and assignment performance. 
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 808 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNmqzbTBmbusrWLekIV:-NNmqzbTBmbusrWLekIW:bj67apr) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L808)):*
> ```Java
>     Map<String, Integer> delegateTaskCountMap =
>         eligibleListOfDelegates.stream().collect(toMap(delegateId -> delegateId, taskCount -> 0));
>     Map<String, List<String>> delegateTaskList = delegateTasks.stream().collect(
> ```

This is not really needed (also, don't use data structure names in the variable names). You can simplify it quite a lot by just mapping tasks to delegateId (which is what you need at the end), then count the number of occurrences (each occurrence is from one task) and sort (you also wouldn't need comparator for equals since order is not guarantee, but also you shouldn't care about breaking that tie). Something like this [untested]



_Suggestion:_
```Java
    final var tasksCount = tasks.stream().collect(groupingBy(DelegateTask::getDelegateId, counting()));
    return tasksCount.entrySet().stream().sorted(Map.Entry.comparingByValue()).map(Map.Entry::getKey).collect(toList());
```

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 794 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNmtmBV9mumGOkb9F0e:-NNn0Ehi3RE-7k31snMQ:b-kyl4be) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L794)):*

Is this done for every task assignment or just e.g. sync tasks (do only sync tasks have first attempt unicast)? I think it might have big impact on mongo. We can try it out, but we should keep an eye on mongo and assignment performance. 
</blockquote></details>

only for NG and under FF for now till we have dedicated DB. In discussion with DBA on adding right index, that mite come in next checkin

mdmilic (changes requested):
> Reviewed 1 of 4 files at r6, 1 of 1 files at r11, 2 of 3 files at r12.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43118)* status: :shipit: complete! ✓

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 770 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn1-6W9Yd-qtX9ShJW:-NNn1-6W9Yd-qtX9ShJX:bi4arx2) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L770)):*
> ```Java
>       List<String> eligibleDelegatesSorted =
>           getEligibleDelegateListOrderedNumberByTaskAssigned(delegateTask, eligibleListOfDelegates);
>       if (!isEmpty(eligibleDelegatesSorted)) {
> ```

This should always have the same size as the input list. I don't think there is a real need for this check, if the input list is empty just short circuit by throwing IllegalState or something similar. You would fail on bellow line anyway. 



_Code snippet:_
```Java
return eligibleListOfDelegates.get(random.nextInt(eligibleListOfDelegates.size()));
```
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 777 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn0cmI1Cjlrx6CmMxk:-NNn0cmI1Cjlrx6CmMxl:bgvyn3m) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L777)):*
> ```Java
>       }
>       // if no task currently running then choose random from the list
>       return eligibleListOfDelegates.get(random.nextInt(eligibleListOfDelegates.size()));
> ```

No real need for this, just always pick first from sorted list
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 802 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn04sa4BdAmzdjFqqQ:-NNn04sa4BdAmzdjFqqR:b-a60sbf) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L802)):*
> ```Java
>                                            .project(DelegateTaskKeys.delegateId, true)
>                                            .asList();
>     // if no task is running currently then return empty list
> ```

You also wouldn't need this check since streams short-circuit on empty collections.

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 808 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNmqzbTBmbusrWLekIV:-NNn0h0S1HRjU9TtiKQA:b-4a5yi5) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L808)):*

This is not really needed (also, don't use data structure names in the variable names). You can simplify it quite a lot by just mapping tasks to delegateId (which is what you need at the end), then count the number of occurrences (each occurrence is from one task) and sort (you also wouldn't need comparator for equals since order is not guarantee, but also you shouldn't care about breaking that tie). Something like this [untested]
</blockquote></details>

we need a comparator, here note we are fetching tasks from DB which is in STARTED state of eligible delegate list in account, then group by delegateId to aggregate count. Then updating the existing eligible delegate list with the count. Note final map should contain all eligible list even with zero tasks running. Also if the value is the same between two delegates then picking random in deciding the order is important as it helps to avoid going to the same delegate. \

I will remove the 'map' from the name

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 770 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn1-6W9Yd-qtX9ShJW:-NNn2buC17rEozQNmHLv:b-uiq2vy) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L770)):*

This should always have the same size as the input list. I don't think there is a real need for this check, if the input list is empty just short circuit by throwing IllegalState or something similar. You would fail on bellow line anyway. 
</blockquote></details>

this is important not to overload one delegate. Note this is existing way of picking first broadcast delegate

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 802 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn04sa4BdAmzdjFqqQ:-NNn2v03B2yVxA12L2Un:bof6711) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L802)):*

You also wouldn't need this check since streams short-circuit on empty collections. 
</blockquote></details>

it can throw exception

mdmilic (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43118)* status: :shipit: complete! ✓

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 770 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn1-6W9Yd-qtX9ShJW:-NNn4bdQ7eBSjA9JFAuz:bi2jdnn) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L770)):*

this is important not to overload one delegate. Note this is existing way of picking first broadcast delegate
</blockquote></details>

This check is only reached with FF enabled. if your sorted list is empty, that means your input list is empty, so once you reach the line in comment you will always throw IndexOutOfBounds
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 777 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn0cmI1Cjlrx6CmMxk:-NNn3xGB1MqKSqf6y9Dg:ba356q2) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L777)):*

this is needed if no task in running state for eligible delegate list
</blockquote></details>

Still not needed, just use whatever sorted list you got and pick first. Sorted list would still have all delegateIds even for delegates with no tasks. 
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 802 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn04sa4BdAmzdjFqqQ:-NNn3WSeE1QnI8oYWpje:b-3epwnu) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L802)):*

it can throw exception
</blockquote></details>

How/Which? You can't have nulls here (your asList() guarantees empty list for no results), streaming empty list would just give you empty collection and skip any operations. 
___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 808 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNmqzbTBmbusrWLekIV:-NNn2TYS4rgebpLTGvJJ:bif7fw) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L808)):*

we need a comparator, here note we are fetching tasks from DB which is in STARTED state of eligible delegate list in account, then group by delegateId to aggregate count. Then updating the existing eligible delegate list with the count. Note final map should contain all eligible list even with zero tasks running. Also if the value is the same between two delegates then picking random in deciding the order is important as it helps to avoid going to the same delegate. \

I will remove the 'map' from the name
</blockquote></details>

I understood the logic, but going to the same delegate doesn't really matter, as they would have the same number of tasks. Also, order is not really guaranteed with the streaming set AFAIK (so if you run this 100 times on the same input, you would get different order for ties). Some collections do guarantee stream order (I think entryset is not one of them), for those you can just always do `.stream().unordered()` and again avoid resolving ties manually and still keep randomness. But again, you really don't need to care about resolving ties here.

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 808 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNmqzbTBmbusrWLekIV:-NNn5O_BCph3xxNdJDLC:bd7q5il) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L808)):*

I understood the logic, but going to the same delegate doesn't really matter, as they would have the same number of tasks. Also, order is not really guaranteed with the streaming set AFAIK (so if you run this 100 times on the same input, you would get different order for ties). Some collections do guarantee stream order (I think entryset is not one of them), for those you can just always do `.stream().unordered()` and again avoid resolving ties manually and still keep randomness. But again, you really don't need to care about resolving ties here. 
</blockquote></details>

it can overload one delegate if not applied random. Note there is the time difference between when delegates gets picked on first broadcast and time when it get acquired and assigned. Without applying random we dont get good result

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 808 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNmqzbTBmbusrWLekIV:-NNn61MdB2nK3SsUOqGV:b-tmoe01) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L808)):*

it can overload one delegate if not applied random. Note there is the time difference between when delegates gets picked on first broadcast and time when it get acquired and assigned. Without applying random we dont get good result
</blockquote></details>

This has been tested all over again and again and it's the best approach to achieve fair distribution

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 770 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn1-6W9Yd-qtX9ShJW:-NNn7DJf5zLdJGtpe1DU:blwylcs) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L770)):*

This check is only reached with FF enabled. if your sorted list is empty, that means your input list is empty, so once you reach the line in comment you will always throw IndexOutOfBounds
</blockquote></details>

sorted listed empty doesnt mean input list is empty here. EligibleDelegateList will have at most one here, otherwise we it will not reach here

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 777 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn0cmI1Cjlrx6CmMxk:-NNn7XhE4UuEVb4mwY9e:b5frsyj) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L777)):*

Still not needed, just use whatever sorted list you got and pick first. Sorted list would still have all delegateIds even for delegates with no tasks. 
</blockquote></details>

we get an empty list when no tasks are running as per the currrent logic. The name may be misleading

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 802 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn04sa4BdAmzdjFqqQ:-NNn7tRj7wIWpGowbU6H:bd8xd10) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L802)):*

How/Which? You can't have nulls here (your asList() guarantees empty list for no results), streaming empty list would just give you empty collection and skip any operations. 
</blockquote></details>

groupingBy i think, not sure. I remember encountered NPE at some point

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 777 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn0cmI1Cjlrx6CmMxk:-NNn8T2V4I_b6jedS95G:bndxcnm) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L777)):*

we get an empty list when no tasks are running as per the currrent logic. The name may be misleading
</blockquote></details>

However it's , we need to apply random on sortedList or eligible list

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 808 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNmqzbTBmbusrWLekIV:-NNnO4Qh5uFZFiztWs_C:bp2mivf) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L808)):*

This has been tested all over again and again and it's the best approach to achieve fair distribution
</blockquote></details>

updated map name.

mdmilic (changes requested):
> Reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43118)* status: 1 change requests. 3. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 808 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNmqzbTBmbusrWLekIV:-NNrNW4X6UFPMKC2osfy:blysvro) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L808)):*

updated map name. 
</blockquote></details>

I think I understand what the problem here is. This method is supposed to be just ordering delegateIds, but instead if delegate doesn't have tasks it gets removed from the list. This is obviously wrong. Updated sample that would fix this is something like this



_Code snippet:_
```Java
final var tasksCount = tasks.stream().collect(groupingBy(DelegateTask::getDelegateId, counting()));
delegates.forEach(delegate -> tasksCount.computeIfAbsent(delegate, key -> 0L));
return tasksCount.entrySet().stream().sorted(Map.Entry.comparingByValue()).map(Map.Entry::getKey).collect(toList());
```

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 770 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn1-6W9Yd-qtX9ShJW:-NOCtQDs5skFtUWfcQDb:bvnrpcm) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L770)):*

sorted listed empty doesnt mean input list is empty here. EligibleDelegateList will have at most one here, otherwise we it will not reach here
</blockquote></details>

updated

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 777 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn0cmI1Cjlrx6CmMxk:-NOCtSuH2JBXmtpJ2vk3:bvnrpcm) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L777)):*

However it's , we need to apply random on sortedList or eligible list
</blockquote></details>

updated

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 802 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNn04sa4BdAmzdjFqqQ:-NOCtUwJ2Mo46eqZ6Vi-:bvnrpcm) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L802)):*

groupingBy i think, not sure. I remember encountered NPE at some point 
</blockquote></details>

updated

jennyjames21:
*[`400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java` line 808 at r12](https://reviewable.io/reviews/harness/harness-core/43118#-NNmqzbTBmbusrWLekIV:-NOCtWjQ4NQ5JpZFPmRS:bvnrpcm) ([raw file](https://github.com/harness/harness-core/blob/cb0e195c4985e788fcc35bbaca009dc1b9064a60/400-rest/src/main/java/software/wings/service/impl/DelegateTaskServiceClassicImpl.java#L808)):*


Issue 1: but still don't random part here, it's important the random should be there to get better result.

Concern 1:  it's unfortunate counting() returns long instead of int as it adds another layer of unnecessary complexity. That's what i would prefer to keep as a list, IMO it's more simpler and I would like to keep it that way. Note Mongo query only returns 2 fields, not the entire document here.


</blockquote></details>

updated

mdmilic (approved):
> Reviewed 1 of 2 files at r14, 1 of 1 files at r15, 2 of 2 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43118)* status: :shipit: complete! ✓

jennyjames21:
trigger ft2

PR #42635, opened 2023-01-16
feat: [Del-5792]: tasklib

xingchi-jin:
## Describe your changes
K8s task separation and containerization
Executorlib implementation that can enable the containerization of tasks

Test Status:
Tested with pre-QA manager + delegate core + task container
image
Screen Shot 2023-02-03 at 4 28 19 PM

## Checklist
- [ ] I've documented the changes in the PR description.
- [ ] I've tested this change either in PR or local environment.
- [ ] If hashcheck failed I followed [the checklist](https://harness.atlassian.net/wiki/spaces/DEL/pages/21016838831/PR+Codebasehash+Check+merge+checklist) and updated this PR description with my answers to make sure I'm not introducing any breaking changes.

## Comment Triggers
<details>
  <summary>Build triggers</summary>

- Feature build: `trigger feature-build`
- Immutable delegate `trigger publish-delegate`
</details>

<details>
  <summary>PR Check triggers</summary>

You can run multiple PR check triggers by comma separating them in a single comment. e.g. `trigger ti0, ti1`

- Compile: `trigger compile`
- CodeformatCheckstyle: `trigger checkstylecodeformat`
    - CodeFormat: `trigger codeformat`
    - Checkstyle: `trigger checkstyle`
- MessageMetadata: `trigger messagecheck`
- File-Permission-Check: `trigger checkpermission`
- Recency: `trigger recency`
- BuildNumberMetadata: `trigger buildnum`
- PMD: `trigger pmd`
- Copyright Check: `trigger copyrightcheck`
- Feature Name Check: `trigger featurenamecheck`
- TI-bootstrap: `trigger ti0`
- TI-bootstrap1: `trigger ti1`
- TI-bootstrap2: `trigger ti2`
- TI-bootstrap3: `trigger ti3`
- TI-bootstrap4: `trigger ti4`
- FunctionalTest1: `trigger ft1`
- FunctionalTest2: `trigger ft2`
- CodeBaseHash: `trigger codebasehash`
- CodeFormatCheckstyle: `trigger checkstylecodeformat`
- SonarScan: `trigger ss`
- GitLeaks: `trigger gitleaks`
- Trigger all Checks: `trigger smartchecks`
</details>

## PR check failures and solutions
https://harness.atlassian.net/wiki/spaces/BT/pages/21106884744/PR+Checks+-+Failures+and+Solutions

[...]

PR #43716, opened 2023-02-08
feat: [del-5792]: Executor library

xingchi-jin:
## Describe your changes
Executorlib implementation that can enable the containerization of tasks

This PR is a split from https://github.com/harness/harness-core/pull/42635

Test Status:
Tested with pre-QA manager + delegate core + task container

## Checklist

## Comment Triggers
<details>

</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- TI-bootstrap: `trigger ti0`
- SonarScan: `trigger ss`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/43716)


mdmilic (changes requested):
> Reviewed 23 of 25 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43716)* status: 1 change requests. 0 of 1 approvals obtained. 

___
*[`delegate/src/main/java/io/harness/delegate/executor/BUILD.bazel` line 15 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpqGAO-7Yn59swNzaa:-NNpqGAO-7Yn59swNzab:bkxenkb) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/BUILD.bazel#L15)):*
> ```Starlark
>     deps = [
>         "//930-delegate-tasks/src/main/java/io/harness/delegate/task/common:module",
>         "//950-delegate-tasks-beans:module",
> ```

shouldn't need (entire) task beans
___
*[`delegate/src/main/java/io/harness/delegate/executor/BUILD.bazel` line 25 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpq7yF74Qc87hQH998:-NNpq7yF74Qc87hQH999:bhak3ar) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/BUILD.bazel#L25)):*
> ```Starlark
>         "//delegate/src/main/java/io/harness/delegate/executor/shutdownhook:module",
>         "//delegate/src/main/java/io/harness/delegate/executor/taskloader:module",
>         "//delegate/src/main/java/io/harness/delegate/taskagent/client/delegate:module",
> ```

You shouldn't need this, agent handles sending response on it's own
___
*[`delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java` line 65 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpr5lz8p11e5kPBfzF:-NNpr5lz8p11e5kPBfzG:b-k32i22) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java#L65)):*
> ```Java
>     DelegateRunnableTask delegateRunnableTask = on(classMap.get(TaskType.valueOf(delegateTaskPackage.getData().getTaskType())))
>             .create(delegateTaskPackage,
>                 /* TBD add stream logger */ null,
> ```

tasks will not send logs directly, task agent redirects tasks std::out to itself and streams that over to the logging service.
___
*[`delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java` line 75 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpqXW-EM6EgP8tstG1:-NNpqXW-EM6EgP8tstG2:b-ujekmi) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java#L75)):*
> ```Java
>   }
> 
>   public Consumer<DelegateTaskResponse> getPostExecutionFunction(
> ```

result of the executor is to write a response to a file. Task Agent will pick it up from there and send it to the delegate. Task agent part can be implemented in a separate PR. 
___
*[`delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java` line 93 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpqS-m6AD6pTATcasP:-NNpqS-m6AD6pTATcasQ:blqtj63) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java#L93)):*
> ```Java
> 
>   // FIXME: We will initialize log service here
>   private BooleanSupplier getPreExecutor() {
> ```

logging will be happening from task agent
___
*[`delegate/src/main/java/io/harness/delegate/executor/bundle/BUILD.bazel` line 18 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNppq5Z4EIgxh0do23B:-NNppq5Z4EIgxh0do23C:bcvo2l6) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/bundle/BUILD.bazel#L18)):*
> ```Starlark
>     deps = [
>         "//930-delegate-tasks/src/main/java/io/harness/delegate/task/common:module",
>         "//950-delegate-tasks-beans:module",
> ```

why entire task beans
___
*[`delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java` line 30 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpyiT10LaBiUmM6P1J:-NNpyco0-SG3SdcDxqqP:b-vtbhn8) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java#L30)):*
> ```Java
>         // Step 2. decode the token, check if the decoded is hex decimal string
>         String decoded = new String(Base64.decodeBase64(delegateToken)).trim();
>         if (isHexDecimalString(decoded)) {
> ```

Why do we need this check? 
If it already passed step 1, then it must be base64, so we decode it and return it. 
Why would we return it encoded base64 version (if decoded is not in hex format) i.e. line 33 `return delegateToken;`



_Code snippet:_
```Java
        if (isHexDecimalString(decoded)) {
```
___
*[`delegate/src/main/java/io/harness/delegate/executor/config/Configuration.java` line 14 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNq-A0X0bkYDTjg5OeW:-NNq-A0X0bkYDTjg5OeX:bbi5yfu) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/config/Configuration.java#L14)):*
> ```Java
> import lombok.ToString;
> import lombok.Value;
> import org.apache.commons.codec.binary.Base64;
> ```

this is probably orphaned import
___
*[`delegate/src/main/java/io/harness/delegate/executor/config/Configuration.java` line 31 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpzqdvA0AQeABwPrEB:-NNpzqdvA0AQeABwPrEC:b-eu0lr8) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/config/Configuration.java#L31)):*
> ```Java
>      */
>     @Builder.Default
>     private boolean shouldSendResponse = false;
> ```

Task executor and task shouldn't know anything about delegate, that's the job of task agent



_Code quote:_
```Java
    private String delegateName;
    /**
     * set to false when testing locally without delegate core
     */
    @Builder.Default
    private boolean shouldSendResponse = false;
    private String delegateHost;
    private int delegatePort;
```
___
*[`delegate/src/main/java/io/harness/delegate/executor/config/ConfigurationProvider.java` line 24 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNq-eu039hhICOs6via:-NNq-eu039hhICOs6vib:bfy69l5) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/config/ConfigurationProvider.java#L24)):*
> ```Java
> 
> @Slf4j
> public class ConfigurationProvider {
> ```

There is probably no point in having executor config in file, just wastes time reading files. It should just be hardcoded. Executor belongs to the task side of things, so task developers (when containerizing) can just configure their module in a way specific to those tasks. Input file location should just be hardcoded, I don't see a reason now to have it configurable, if we do have a need we can implement it in the future. All other params belong to the task agent config
___
*[`delegate/src/main/java/io/harness/delegate/executor/serviceproviders/BUILD.bazel` line 16 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNq11TOAbsl0VsUG2w4:-NNq11TOAbsl0VsUG2w5:b-s2wd5n) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/serviceproviders/BUILD.bazel#L16)):*
> ```Starlark
>         "//920-delegate-agent-beans:module",
>         "//930-delegate-tasks:module",
>         "//950-delegate-tasks-beans:module",
> ```

Can you re-run build cleaner for this. This module depends on so many things, but contains only noop implementations



_Code quote:_
```Starlark
        "//920-delegate-agent-beans:module",
        "//930-delegate-tasks:module",
        "//950-delegate-tasks-beans:module",
        "//957-cg-beans/src/main/java/io/harness/cvng/beans/cvnglog:module",
        "//957-cg-beans/src/main/java/software/wings/verification:module",
        "//957-cg-beans:module",
        "//959-file-service-commons/src/main/java/io/harness/delegate/beans:module",
```
___
*[`delegate/src/main/java/io/harness/delegate/executor/taskloader/TaskPackageReader.java` line 30 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNq1Y_XCxEiuyl2lo6u:-NNq1Y_XCxEiuyl2lo6v:b-twfvce) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/taskloader/TaskPackageReader.java#L30)):*
> ```Java
>     final File taskDataFile = new File(taskDataPath);
>     try {
>       var data = FileUtils.readFileToByteArray(taskDataFile);
> ```

I think standard `java.nio`provides something like `Files.readAllBytes` which you could use instead. I see you depend on apache commons on few places just for reading files like this. I think that dep would still end up in the final binary transitively somehow, but we should probably try to keep it clean. 
___
*[`delegate/src/main/java/io/harness/delegate/taskagent/client/delegate/DelegateCoreClient.java` line 25 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNq29lh9IkzdBsmyjNr:-NNq29lh9IkzdBsmyjNs:b-hrj1ek) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/taskagent/client/delegate/DelegateCoreClient.java#L25)):*
> ```Java
>   @KryoRequest
>   Call<Void> taskResponse(@Path("taskId") final String taskId,
>                           @Body final DelegateTaskResponse delegateTaskResponse);
> ```

checkstyle complains about this because interfaces are implicitly final, I also like making it explicit though :)

mdmilic (changes requested):
> Reviewed 5 of 7 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43716)* status: 1 change requests. 0 of 1 approvals obtained. 

___
*[`delegate/src/main/java/io/harness/delegate/executor/BUILD.bazel` line 25 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpq7yF74Qc87hQH998:-NOJsvwf9pCHhyehfAkB:br8j859) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/BUILD.bazel#L25)):*

You shouldn't need this, agent handles sending response on it's own
</blockquote></details>

executor lib should not depend on task agent. Should be the other way around, also this would create circular dep
___
*[`delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java` line 75 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpqXW-EM6EgP8tstG1:-NOJpmP32j5zHQbTfmLv:b-ho0775) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java#L75)):*

I think we had agreed on for Harness tasks, we will have everything in process. When the feature of community task comes, we can change this part of code to write files 
but let me know, we should discuss on this
</blockquote></details>

Yes, everything should be in process, but the responsibilities remain the same, executor lib (task) provides task response to the task agent, task agent sends it to delegate. Executor lib should not assume responsibilities of the task agent. The only difference is that you don't need to care about inter process communication between task agent and task (executor lib in this case), you just return the response from method invocation. I.e. this is how it should look like (simplified):
Also, don't worry about task agent in this PR, changes there should go separately. 



_Code snippet:_
```Java
// Pseudo Java - current state
public TaskAgen {
  public main(...) {
    TaskResponse response = taskExecutor.runTask(...);
    Endpoint endpoint = serviceDiscovery.getDelegateCoreServiceEndpoint();
    delegateCoreClient = clientFactory.createClient(endpoint)
    delegateCoreClient.sendResponse(wrapInProto(response));
  }
}

// Pseudo Java - final state
public TaskAgen {
  public main(...) {
    String runCommand = config.getRunCommand();
    Process taskProcess = ProcessRunner.runProcess(runCommand); // Here you can set timeout, redirect output, process destroyer etc...
    await tasKProcess
    Endpoint endpoint = serviceDiscovery.getDelegateCoreServiceEndpoint();
    delegateCoreClient = clientFactory.createClient(endpoint)
    byte[] response = readResponseFile(config.getResponsePath());
    delegateCoreClient.sendResponse(wrapInProto(response));
  }
}
```
___
*[`delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java` line 30 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpyiT10LaBiUmM6P1J:-NOJnD_x1TwgCa-nlr_5:bnx8xpn) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java#L30)):*

because we introduced a new feature this release, where we base64 encode the token. 
the old token is still a valid base64 string, so just by step1 it doesn't tell if this is a new token or old 
</blockquote></details>

Ok, I think the logic is a bit backwards then, you decode first and then check if you should even be decoding, if no you throw away the result and use old value. So would below work instead? With this change in case of old token you don't have unnecessary decoding which you throw away. 



_Code snippet:_
```Java
        // Step 2. if hex that means old token
        if (isHexDecimalString(delegateToken)) {
            return delegateToken;
        }
        // Step 3. if not hex, then we need to decode it
        return new String(Base64.decodeBase64(delegateToken)).trim();
```
___
*[`delegate/src/main/java/io/harness/delegate/executor/config/Configuration.java` line 31 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpzqdvA0AQeABwPrEB:-NOJk3KQ8sKwwR4zI2vr:b-jm7qvm) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/config/Configuration.java#L31)):*

Done.

delegateHost and port is a replacement of the current way, where they are set in environment variables. basically the same thing. But it's better to put everything in config file
</blockquote></details>

There is no need for replacement. Task Agent already handles this, executor lib (task) should not know anything about delegate and should not send responses to it. Executor lib provides its response to task agent which handles the rest. Also executor lib (task) should not depend on task agent (it's the other way around right now so this also creates circular dep)

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43716)* status: 1 change requests. 0 of 1 approvals obtained. 

___
*[`delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java` line 30 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpyiT10LaBiUmM6P1J:-NOJuQXq8xYDibX2drf8:bzbe87p) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java#L30)):*

Ok, I think the logic is a bit backwards then, you decode first and then check if you should even be decoding, if no you throw away the result and use old value. So would below work instead? With this change in case of old token you don't have unnecessary decoding which you throw away. 
</blockquote></details>

Also, the check to see if it's already base64 is unnecessary since it's always true (for old and new token)



_Code snippet:_
```Java
        if (!Base64.isBase64(delegateToken)) {
            return delegateToken;
        }
```

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43716)* status: 1 change requests. 0 of 1 approvals obtained. 

___
*[`delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java` line 30 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpyiT10LaBiUmM6P1J:-NOJvTN4Am8j5OVdWuqk:b-czycmq) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java#L30)):*

Also, the check to see if it's already base64 is unnecessary since it's always true (for old and new token)
</blockquote></details>

Last thing I wanted to mention, not sure how often this code gets executed, but using regex (even though statically compiled), might be too much if this gets used often. We know that decoded token needs to be 32 chars which is also in your regex (encoded is around 40 or so), so instead of regex we can just do `delegateToken.length() == 32` and don't care about verifying that it's hex, that property is really not important for us to decide if we need to decode the token or not.

xingchi-jin (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43716)* status: 1 change requests. 3. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java` line 75 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpqXW-EM6EgP8tstG1:-NOMFUpw6HHm4XGaCZLc:b1t6760) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java#L75)):*

Yes, everything should be in process, but the responsibilities remain the same, executor lib (task) provides task response to the task agent, task agent sends it to delegate. Executor lib should not assume responsibilities of the task agent. The only difference is that you don't need to care about inter process communication between task agent and task (executor lib in this case), you just return the response from method invocation. I.e. this is how it should look like (simplified):
Also, don't worry about task agent in this PR, changes there should go separately. 

</blockquote></details>

Task agent should not have the main function at this stage, the k8s ng task does. 
executor lib currently uses some task agent codes to send response. When separating out task agent to be a process, executor lib just replace the codes of sending responses to writing to a file. 

no need to struggle too much at code detail level. 
___
*[`delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java` line 30 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpyiT10LaBiUmM6P1J:-NOMHSpk2eX1QgaUmPpn:brpr10n) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java#L30)):*

Last thing I wanted to mention, not sure how often this code gets executed, but using regex (even though statically compiled), might be too much if this gets used often. We know that decoded token needs to be 32 chars which is also in your regex (encoded is around 40 or so), so instead of regex we can just do `delegateToken.length() == 32` and don't care about verifying that it's hex, that property is really not important for us to decide if we need to decode the token or not. 
</blockquote></details>

the base64 encoded token (new token) is HexDecimalString  as well. Then in the approach you mentioned, it will not be decoded 
base64 decode is not compute heavy, this operations runs only once at starting time.

Instead of solving this problem, I think we shouldn't use delegate token at all in task container, but it can be next task to fix it 
___
*[`delegate/src/main/java/io/harness/delegate/executor/config/Configuration.java` line 31 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpzqdvA0AQeABwPrEB:-NOMHy98AeIhQ0FWvTle:b-9f1myz) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/config/Configuration.java#L31)):*

There is no need for replacement. Task Agent already handles this, executor lib (task) should not know anything about delegate and should not send responses to it. Executor lib provides its response to task agent which handles the rest. Also executor lib (task) should not depend on task agent (it's the other way around right now so this also creates circular dep)
</blockquote></details>

again, it's abt the question what is task agent Vs what is task, currently the executor lib directly sends the response back.
once we have task agent, the task agent will load these two envs [host and port], and we will remove them from executor lib. 
does it sounds good to you ?

xingchi-jin (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43716)* status: 1 change requests. 4. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`delegate/src/main/java/io/harness/delegate/executor/BUILD.bazel` line 25 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpq7yF74Qc87hQH998:-NOMJ0if0apmEMAKryiM:b-n1f7p5) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/BUILD.bazel#L25)):*

executor lib should not depend on task agent. Should be the other way around, also this would create circular dep
</blockquote></details>

executor will use task agent as a lib at step 1 
___
*[`delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java` line 65 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpr5lz8p11e5kPBfzF:-NOMIqCk2OBW1UUF8O8a:b-emxjhv) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/TaskFactory.java#L65)):*

tasks will not send logs directly, task agent redirects tasks std::out to itself and streams that over to the logging service.
</blockquote></details>

yes, that's step 3 as we discussed yesterday

mdmilic (approved):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43716)* status: Some of the checks are too old. 

___
*[`delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java` line 30 at r2](https://reviewable.io/reviews/harness/harness-core/43716#-NNpyiT10LaBiUmM6P1J:-NOq-rpg0eA_lj5quf8P:b-m1pntf) ([raw file](https://github.com/harness/harness-core/blob/574c22d3bd8eca55101aaf0c9780b4f9003ffea6/delegate/src/main/java/io/harness/delegate/executor/common/DelegateTokenUtils.java#L30)):*
> the base64 encoded token (new token) is HexDecimalString  as well
Yes, but it would fail the check since it has != 32 chars which your regex requires (and I think probably instead of regex we could do 'delegateToken.length() == 32'). This was more general question since I know the same exact code exists on delegate.

xingchi-jin:
trigger checkstylecodeformat

xingchi-jin:
trigger checkstylecodeformat

mdmilic (approved):
:lgtm:

> Reviewed 5 of 7 files at r6, 19 of 19 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/43716#-:-NOzo2EbEGyDJv0-4J-L:bnfp4nl)* status: :shipit: complete! ✓

PR #44857, opened 2023-03-07
fix:[CDS-53786]: Add support to query specific tag from quay registry.

Fernando-Dourado:
## Describe your changes
- More details when the final PR comes.
- Btw, we can use the same API to retrieve the list of tags increasing the customer experience.

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/44857)


Fernando-Dourado:
Fixed in https://harness.atlassian.net/browse/CDS-54085

PR #45317, opened 2023-03-17
fix:[CDS-56765]: Ignore and log warn when audited entities do not have userId reference

Fernando-Dourado:
## Describe your changes
- Ignore and log warn when audited entities do not have userId reference

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/45317)

### DIFF
--- a/400-rest/src/main/java/software/wings/service/impl/AuditServiceImpl.java
+++ b/400-rest/src/main/java/software/wings/service/impl/AuditServiceImpl.java
@@ -477,11 +477,12 @@ public class AuditServiceImpl implements AuditService {
     }
   }
 
-  private AuditHeader getAuditHeaderById(@NotNull String Id) {
-    return wingsPersistence.createQuery(AuditHeader.class).filter(AuditHeader.ID_KEY2, Id).get();
+  private AuditHeader getAuditHeaderById(@NotNull String id) {
+    return wingsPersistence.createQuery(AuditHeader.class).filter(AuditHeader.ID_KEY2, id).get();
   }
 
-  private <T> void addDetails(String accountId, T entity, String auditHeaderId, Type type) {
+  @VisibleForTesting
+  <T> void addDetails(String accountId, T entity, String auditHeaderId, Type type) {
     if (auditHeaderId == null) {
       return;
     }
@@ -494,7 +495,11 @@ public class AuditServiceImpl implements AuditService {
     } else if (entity instanceof ApiKeyEntry && type.equals(Type.INVOKED)) {
       entityMetadataHelper.addAPIKeyDetails(accountId, entity, header);
     } else if (header.getCreatedBy() != null) {
-      entityMetadataHelper.addUserDetails(accountId, entity, header);
+      if (isNotEmpty(header.getCreatedBy().getUuid())) {
+        entityMetadataHelper.addUserDetails(accountId, entity, header);
+      } else {
+        log.warn("Unable to audit entity due to missing userId [entity={},type={}]", entity.getClass(), type);
+      }
     }
   }
 
--- a/400-rest/src/test/java/software/wings/service/impl/AuditServiceImplTest.java
+++ b/400-rest/src/test/java/software/wings/service/impl/AuditServiceImplTest.java
@@ -7,8 +7,10 @@
 
 package software.wings.service.impl;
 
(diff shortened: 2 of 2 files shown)


Fernando-Dourado (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/45317)* status: 2 file(s) to review. 1 change requests. 1. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/AuditServiceImpl.java` line 499 at r1](https://reviewable.io/reviews/harness/harness-core/45317#-NQl5r6q8on5AeRMhAF6:-NQl6jZDAermUF9AC3Ns:b-8zd2nq) ([raw file](https://github.com/harness/harness-core/blob/1bf816fbcd43cbe2454ae5cfca24787f9b4c0c0d/400-rest/src/main/java/software/wings/service/impl/AuditServiceImpl.java#L499)):*
<details><summary><i>Previously, abhinav-harness (Abhinav Singh) wrote…</i></summary><blockquote>

Can we make it like  - dont want to make any big change 
```      
      entityMetadataHelper.addUserDetails(accountId, entity, header);
if (isEmpty(header.getCreatedBy().getUuid())) {
        log.warn("Unable to audit entity due to missing userId [entity={},type={}]", entity.getClass(), type);
      }```
```

</blockquote></details>

But, doing that way the exception keeps throwing every time the code is executed.

Another option is to identify when the execution comes from API or not, it simply just evaluates if the header has X-API-Key set. What is your opinion on that?

abhinav-harness:
*[`400-rest/src/main/java/software/wings/service/impl/AuditServiceImpl.java` line 499 at r1](https://reviewable.io/reviews/harness/harness-core/45317#-NQl5r6q8on5AeRMhAF6:-NQl8NGe0y1K-JUaLx2q:b-q4l277) ([raw file](https://github.com/harness/harness-core/blob/1bf816fbcd43cbe2454ae5cfca24787f9b4c0c0d/400-rest/src/main/java/software/wings/service/impl/AuditServiceImpl.java#L499)):*
<details><summary><i>Previously, Fernando-Dourado (Fernando Dourado) wrote…</i></summary><blockquote>

But, doing that way the exception keeps throwing every time the code is executed.

Another option is to identify when the execution comes from API or not, it simply just evaluates if the header has X-API-Key set. What is your opinion on that?
</blockquote></details>

ok we are getting exception in such cases - can we keep like this (the comment) for now and identify root cause and fix. (let error logs come for more time). ?

Fernando-Dourado (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/45317)* status: 2 file(s) to review. 1 change requests. 1. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/AuditServiceImpl.java` line 499 at r1](https://reviewable.io/reviews/harness/harness-core/45317#-NQl5r6q8on5AeRMhAF6:-NQlJfGY9yehpnjTSrgS:b7rcobr) ([raw file](https://github.com/harness/harness-core/blob/1bf816fbcd43cbe2454ae5cfca24787f9b4c0c0d/400-rest/src/main/java/software/wings/service/impl/AuditServiceImpl.java#L499)):*
<details><summary><i>Previously, abhinav-harness (Abhinav Singh) wrote…</i></summary><blockquote>

ok we are getting exception in such cases - can we keep like this (the comment) for now and identify root cause and fix. (let error logs come for more time). ?
</blockquote></details>

The root cause is the operation starting from API execution.

If you do the same using the UI the exception is not thrown because the audited entity has the information of the user doing the action. When it comes from the API we don't know who is the user, we only know the user group.

As I said it is simple to identify and add a condition to don't audit again when it comes from API call.

yogesh-chauhan (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/45317)* status: 2 file(s) to review. 1 change requests. 1. unresolved discussions. 1 of 2 approvals obtained.

Fernando-Dourado:
We are going to split this change into two different pieces to reduce the blasting radio of a side effect.

1. Fix the root cause of delegate tag updates
2. Fix the root cause of CaC updates

PR #45964, opened 2023-03-31
fix:[CDS-54879]: FF to verify appId inside appPermissions of an userGroup

Fernando-Dourado:
## Describe your changes
A user group could still have a reference to a deleted appId when the de-reference/cleanup process doesn't reach the user group entity. It's behind an FF as the performance can be affected when many appIds are related to the user group.

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/45964)


abhinav-harness (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/45964)* status: 3 file(s) to review. 1 of 2 approvals obtained.

PR #46766, opened 2023-04-21
[feat]: [PIE-9801]: Infra Rollback Case Fix

namanvermah:
Recording Link: https://harness.slack.com/archives/C03STPGBGGH/p1682068294520149?thread_ts=1682057753.172999&cid=C03STPGBGGH
The `steps` section is present under both execution and provisioner. In order to decide which `steps` will actually point to the combined rollback steps, the step that fails needs to publish the failure level at stage level.
These are the cases and corresponding solutions:
1. Some step under provisioner steps fails. Then in the rollback stage, the combined rollback steps should start after the provisioner steps
2. Some step under execution steps fails. Then in the rollback stage, the combined rollback steps should start after the execution steps
3. When pipeline rollback reaches a successful stage, the rollback steps should start after the execution steps.

The new advisor on CDSteps for rollback mode checks the Levels from the sweeping output that was published by the failed step, and checks its own level. For case 3 above, there won't be any sweeping output (because it is only published by the failed step), hence a check for current level is required.

Q. Why is the sweeping output at the stage level and not pipeline level?
A. This is because if two parallel stages fail, one at provisioner, another at execution, then their sweeping outputs should not interfere.

This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/46766)


sahilhindwani-harness (dismissed):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/46766)* status: 9 file(s) to review. 1 of 2 approvals obtained.

yogesh-chauhan (approved):
> Reviewed 5 of 9 files at r1, 6 of 6 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/46766)* status: 7 file(s) to review. 1 of 2 approvals obtained.

sahilhindwani-harness (approved):
> Reviewed 5 of 6 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/46766)* status: 1 file(s) to review.

PR #46870, opened 2023-04-24
[feat]: [PIE-9682]: Provide original execution stage ID to rollback steps

namanvermah:
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/46870)


namanvermah:
approach discarded because for matrix one plan node can have multiple node executions

PR #47899, opened 2023-05-16
feat: [CDS-68960]: Changes in Polling Documents subscription for multi artifact Triggers

vcalasansh:
## Describe your changes
This PR implements the required changes in Polling Documents subscription logic, in order to support multi-artifact Triggers.

**Highlights**:
- Added support for pollingRepository delete (or remove signature from existing pollingDoc) when receiving only the accountId and signature (without the pollingDocId). This is useful so we don't have to keep track of both signature and pollingDocId in MultiArtifact Trigger's metadata when unsubscribing from pollingDocuments.
- Had to change `PollingItemGenerator`'s method signature `getBaseInitializedPollingItem(
      NGTriggerEntity ngTriggerEntity)` to `getBaseInitializedPollingItem(
      NGTriggerEntity ngTriggerEntity, BuildTriggerOpsData buildTriggerOpsData)`, because for multiArtifact Triggers the ngTriggerEntity is not enough for knowing which pollingDocument we are supposed to generate (since this trigger can contain more than one). So we are fetching the necessary polling Info from `buildTriggerOpsData` in MultiArtifact Trigger case. This has generated a one-line change in all the classes that implement `PollingItemGenerator`, just to account for this signature change.
- A slightly more generalized logic for subscription/unsubscription was implemented in `executePollingSubscriptionChanges`. This is necessary for MultiArtifact Triggers because the number of pollingItems tied to the trigger can change (think of the case where we are updating a trigger that was listening on 2 sources to listen in 3 sources instead). So the unsubscription logic has to be carried out separately, because there is no necessary similarity between the previous pollingItems and the new ones.
-  The method which fetches trigger's to be executed based in the signature received by ng-manager (when ng-manager starts the trigger execution flow after receiving a new tag for a pollingDoc) was changed to also match signatures present in the new `multiBuildMetadata` field of trigger's metadata. This new field is just a list of the previously existing `buildMetadata`, since for multiArtifact trigger's we can have more than one pollingDocument.

**Tests Conducted**:
- Trigger creation/deletion. Subscription / Unsubscription working ok.
- Trigger update with different artifact sources to listen to (possibly different number of artifact sources too). All ok.
- Trigger enable/disable. All good.
- Trigger execution. Working as expected.

## Checklist
- [x] If hashcheck failed I followed [the checklist](https://harness.atlassian.net/wiki/spaces/DEL/pages/21016838831/PR+Codebasehash+Check+merge+checklist) and updated this PR description with my answers to make sure I'm not introducing any breaking changes.

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/47899)


vcalasansh:
trigger uts

vcalasansh:
trigger uts

vcalasansh:
trigger uts

125-cd-nextgen/modules/common/core/src/main/java/io/harness/repositories/polling/PollingRepositoryCustomImpl.java
```diff
@@ -86,13 +86,15 @@ public PollingDocument deleteDocumentIfOnlySubscriber(String accountId, String o
...
-                                              .and(PollingDocumentKeys.uuid)
-                                              .is(pollingDocId)
-                                              .and(PollingDocumentKeys.signatures)
-                                              .is(signatures));
+    Criteria criteria = new Criteria()
+                            .and(PollingDocumentKeys.accountId)
+                            .is(accountId)
+                            .and(PollingDocumentKeys.signatures)
+                            .is(signatures);
+    if (pollingDocId != null) {
```
sribalij, line 94: When can this be null?

125-cd-nextgen/modules/common/core/src/main/java/io/harness/repositories/polling/PollingRepositoryCustomImpl.java
```diff
@@ -109,7 +111,17 @@ public PollingDocument deleteSubscribersFromExistingPollingDoc(String accountId,
   public PollingDocument removeSubscribersFromExistingPollingDoc(
       String accountId, String uuId, List<String> signatures) {
     Object[] signatureList = signatures.toArray();
-    Query query = getQuery(accountId, uuId);
+    Query query;
+    if (uuId == null) {
```
sribalij, line 115: Can you please elaborate on when uuid will be null?

pipeline-service/modules/ng-triggers/src/main/java/io/harness/ngtriggers/service/impl/NGTriggerServiceImpl.java
```diff
@@ -289,6 +303,89 @@ && executePollingUnSubscription(ngTriggerEntity, pollingItemBytes).equals(Boolea
     }
   }
 
+  private void executePollingSubscriptionChanges(NGTriggerEntity ngTriggerEntity, boolean isUpdate) {
+    // This method is a generalization of `subscribePolling` that we created for handling MultiRegionArtifact Triggers.
+    // We intend to replace `subscribePolling` by this method once we are confident it will introduce no issues.
+    List<PollingItem> pollingItems = pollingSubscriptionHelper.generatePollingItems(ngTriggerEntity);
+
```
sribalij, line 310: Need to find a easy way to map things. 
Handle it for each type seperately in the new method and then phase out the old one?

pipeline-service/modules/ng-triggers/src/main/java/io/harness/ngtriggers/buildtriggers/helpers/BuildTriggerHelper.java
```diff
@@ -242,6 +243,29 @@ public BuildTriggerOpsData generateBuildTriggerOpsDataForArtifact(TriggerDetails
         .build();
   }
 
+  public List<BuildTriggerOpsData> generateBuildTriggerOpsDataForMultiArtifact(TriggerDetails triggerDetails)
+      throws IOException {
+    List<BuildTriggerOpsData> buildTriggerOpsData = new ArrayList<>();
```
sribalij, line 248: Have a few questions around this. Will sync up with you offline

vcalasansh: Ok, thanks!

pipeline-service/modules/ng-triggers/src/main/java/io/harness/ngtriggers/service/impl/NGTriggerServiceImpl.java
vcalasansh: Yes, this method `executePollingSubscriptionChanges` is a generalization of `subscribePolling` - It does the same thing but works for multi-artifact triggers too. The idea is to phase the old method out when we remove the FF for multi-region artifact triggers. 

Previously, all artifact triggers were attached to only one pollingDoc, so the flow was simple - we only need to unsubscribe from one pollingDoc and subscribe to another. The generalization in this new method is that it works for multi-artifacts case too. The difference in multi-artifact triggers is that the number of pollingDocs to unsubscribe can be different of the number of pollingDocs to subscribe (think like you had a trigger listening for US+EU and then update it with US+EU+APAC). This is the generalization implemented in this method.

125-cd-nextgen/modules/common/core/src/main/java/io/harness/repositories/polling/PollingRepositoryCustomImpl.java
vcalasansh: It is because of the update flow for multi-region triggers:
-> **Summary of the old flow**:
- Previously, all artifact triggers were attached to only one pollingDoc and one signature, so the flow was simple - we only need to unsubscribe from one pollingDoc and subscribe to another one.
- When updating a trigger for single artifact, we fetch the pollingDocId from the previously existing data in trigger's DB (metadata.buildMetadata). Then we ask ng-manager to unsubscribe the trigger's signature from that pollingDocId.
-> **Differences with multi-artifact trigger**:
- On multi-artifact triggers, we now have a list of `buildMetadata`, so there is no easy way to do get the pollingDocIds to send for unsubscription.
- So, instead, we only send the trigger signatures and accountId for unsubscription. Which, anyway, is already enough information for ng-manager to handle the unsubscription.

We can discuss more about this offline and I can show you the whole flow.

125-cd-nextgen/modules/common/core/src/main/java/io/harness/repositories/polling/PollingRepositoryCustomImpl.java
vcalasansh: Same as above.

pipeline-service/modules/ng-triggers/src/main/java/io/harness/ngtriggers/service/impl/NGTriggerServiceImpl.java
```diff
@@ -289,6 +303,89 @@ && executePollingUnSubscription(ngTriggerEntity, pollingItemBytes).equals(Boolea
...
+    try {
+      if (isEmpty(pollingItems)) {
+        throw new InvalidRequestException("Cannot subscribe polling for empty pollingItems");
+      }
+      boolean shouldSubscribe = ngTriggerEntity.getEnabled()
+          && !(isWebhookGitPollingEnabled(ngTriggerEntity)
```
sribalij, line 316: Let us refactor this into another method as discussed in the call.

pipeline-service/modules/ng-triggers/src/main/java/io/harness/ngtriggers/service/impl/NGTriggerServiceImpl.java
```diff
@@ -241,17 +244,27 @@ private void registerPollingAsync(NGTriggerEntity ngTriggerEntity, boolean isUpd
     }
 
     // Polling not required for other trigger types
-    if (ngTriggerEntity.getType() != ARTIFACT && ngTriggerEntity.getType() != MANIFEST) {
+    if (ngTriggerEntity.getType() != ARTIFACT && ngTriggerEntity.getType() != MULTI_ARTIFACT
```
sribalij, line 247: Pls add some short comments if possible as discussed

vcalasansh:
trigger review

PR #48208, opened 2023-05-23
[fix]: [CDS-69962]: Include skipped stages in list of PRB stages

namanvermah:
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/48208)


namanvermah:
better to enhance yaml filtering rather than get all skipped stages

PR #49060, opened 2023-06-09
feat: [PL-39351]: Support delegate capacity handling

Raghav-Murali:
## Describe your changes

This PR adds support to register delegate task capacity with the manager. This allows the manager to use the task capacity to determine whether to assign a task to the delegate or to queue it.

Local testing carried out on below -

1. Configure DELEGATE_TASK_CAPACITY to 2 and execute 6 tasks in parallel. Only 2 tasks were being executed at a time.
2. Don't configure DELEGATE_TASK_CAPACITY, executes all 6 tasks in parallel.

## Checklist

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/49060)


mdmilic (commented):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49060)* status: 6 file(s) to review. 0 of 2 approvals obtained. Waiting on team(s): del-team. 

___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 1170 at r1](https://reviewable.io/reviews/harness/harness-core/49060#-NXirALeAwL8UkaD0Tjw:-NXirALeAwL8UkaD0Tjx:b-ad7yeb) ([raw file](https://github.com/harness/harness-core/blob/f505437c4daa154a35043b04fd9900016abae8e2/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L1170)):*
> ```Java
>   private int getDelegateTaskCapacity() {
>     // By default, set the task capacity to 10.
>     int taskCapacity = 10;
> ```

This default is quite low. Delegate was so far able to do 400 tasks in parallel. 
___
*[`400-rest/src/main/resources/delegatetemplates/common/delegate-environment.ftl` line 69 at r1](https://reviewable.io/reviews/harness/harness-core/49060#-NXirqUO0AZjPJSou5i8:-NXirqUO0AZjPJSou5i9:b-7q6jge) ([raw file](https://github.com/harness/harness-core/blob/f505437c4daa154a35043b04fd9900016abae8e2/400-rest/src/main/resources/delegatetemplates/common/delegate-environment.ftl#L69)):*
> ```FreeMarker
>           value: "${dynamicHandlingOfRequestEnabled}"
>         - name: DELEGATE_TASK_CAPACITY
>           value: "${delegateTaskCapacity}"
> ```

This doesn't seem to be set from anywhere.

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49060)* status: 10 file(s) to review. 2. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 1170 at r1](https://reviewable.io/reviews/harness/harness-core/49060#-NXirALeAwL8UkaD0Tjw:-NXlbsYQ4hibNODWvKOf:b-9vath9) ([raw file](https://github.com/harness/harness-core/blob/f505437c4daa154a35043b04fd9900016abae8e2/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L1170)):*

This default is quite low. Delegate was so far able to do 400 tasks in parallel. 
</blockquote></details>

Actually I moved the delegate task capacity handling to delegate instead of manager. So I removed this part. Now delegate will check its current task capacity and if it is more than the customer configured capacity then it won't acquire any new task from the Manager. 
___
*[`400-rest/src/main/resources/delegatetemplates/common/delegate-environment.ftl` line 69 at r1](https://reviewable.io/reviews/harness/harness-core/49060#-NXirqUO0AZjPJSou5i8:-NXlcm8P9PssKUaJZ0_l:b-l8ibuq) ([raw file](https://github.com/harness/harness-core/blob/f505437c4daa154a35043b04fd9900016abae8e2/400-rest/src/main/resources/delegatetemplates/common/delegate-environment.ftl#L69)):*

This doesn't seem to be set from anywhere. 
</blockquote></details>

Updating this logic too. Will use 2 env variables controlled by a FF.

400-rest/src/main/java/io/harness/iterator/FailDelegateTaskIteratorHelper.java
```diff
@@ -93,7 +93,7 @@ public void markLongQueuedTasksAsFailed(DelegateTask delegateTask, boolean isDel
   public void markNotAcquiredAfterMultipleBroadcastAsFailed(
       DelegateTask delegateTask, boolean isDelegateTaskMigrationEnabled) {
     long now = clock.millis();
-    if (delegateTask.getStatus().equals(QUEUED) && delegateTask.getBroadcastRound() <= MAX_BROADCAST_ROUND
+    if (delegateTask.getStatus().equals(QUEUED) && delegateTask.getBroadcastRound() >= MAX_BROADCAST_ROUND
```
xingchi-jin, line 96: is this a bug fix ?

260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java
```diff
@@ -315,6 +315,8 @@ public class DelegateAgentServiceImpl implements DelegateAgentService {
       : DEFAULT_MAX_THRESHOLD;
   private final boolean dynamicRequestHandling = isNotBlank(System.getenv().get("DYNAMIC_REQUEST_HANDLING"))
       && Boolean.parseBoolean(System.getenv().get("DYNAMIC_REQUEST_HANDLING"));
+  private final String checkDelegateCapacity = System.getenv().get("CHECK_DELEGATE_TASK_CAPACITY");
```
xingchi-jin, line 318: use boolean for "checkDelegateCapacity"

260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java
```diff
@@ -2034,6 +2046,13 @@ private void dispatchDelegateTask(DelegateTaskEvent delegateTaskEvent) {
           return;
         }
 
+        // Check the current delegate task capacity and try to acquire the task
+        // only if below the capacity - controlled by feature flag.
+        if ((checkDelegateCapacity.equalsIgnoreCase("true"))
+            && (maxExecutingTasksCount.get() >= getDelegateTaskCapacity())) {
```
xingchi-jin, line 2052: log.info

400-rest/src/main/resources/delegatetemplates/delegate.sh.ftl
```diff
@@ -109,6 +109,8 @@ echo "delegateToken: ${delegateToken}" >> config-delegate.yml
 echo "delegateToken: ${accountSecret}" >> config-delegate.yml
 </#if>
 echo "dynamicHandlingOfRequestEnabled: ${dynamicHandlingOfRequestEnabled}" >> config-delegate.yml
+echo "checkDelegateTaskCapacity: ${checkDelegateTaskCapacity}" >> config-delegate.yml
```
xingchi-jin, line 112: i think just need one approach to pass the config. either using env or using config-delegate.yml

260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java
```diff
@@ -1163,6 +1165,16 @@ private String registerDelegate(final DelegateParamsBuilder builder) {
     return null;
   }
 
+  private int getDelegateTaskCapacity() {
```
raghuAtWings, line 1168: Instead of two env variables we can just have one, the task count and treat zero has infinity
Also this implementation will read the env variable and parse the number for every task. You should read it once at the initialization of delegate and use it instead of reading it for every task

400-rest/src/main/java/io/harness/iterator/FailDelegateTaskIteratorHelper.java
raghuAtWings: Don’t understand this change

400-rest/src/main/resources/delegatetemplates/delegate.sh.ftl
raghuAtWings: Right, there is no reason to put it in config
Yaml

raghuAtWings (changes requested):

260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java
gn-harness: +1

260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java
gn-harness, line 2052: Did we not decide to use currentlyExecutingTasks map?

400-rest/src/main/java/software/wings/app/DelegateQueueTask.java
```diff
@@ -151,7 +152,11 @@ protected void rebroadcastUnassignedTasks(boolean isMigrationEnabled) {
         // if all delegates got one round of rebroadcast, then increase broadcast interval & broadcastRound
         if (alreadyTriedDelegates.containsAll(delegateTask.getEligibleToExecuteDelegateIds())) {
           alreadyTriedDelegates.clear();
-          broadcastRoundCount++;
+          if (!featureFlagService.isEnabled(DELEGATE_TASK_CAPACITY_CHECK, delegateTask.getAccountId())) {
+            // Do not increment the broadcast round count for account(s)
+            // with this FF enabled so the task is rebroadcast until task expiry.
+            broadcastRoundCount++;
+          }
```
gn-harness, line 159: Don't we want an upper bound on this?

400-rest/src/main/resources/delegatetemplates/delegate.sh.ftl
xingchi-jin: @Raghav-Murali  i think @raghuAtWings 's point here is, if the config is directly read from ENV using the java code, instead of using config, then the change can easily apply to both imm and legacy delegate. 
I think it's a good approach to take.

400-rest/src/main/resources/delegatetemplates/common/delegate-environment.ftl
```diff
@@ -65,6 +65,10 @@
           value: ""
         - name: DYNAMIC_REQUEST_HANDLING
           value: "${dynamicHandlingOfRequestEnabled}"
+        - name: CHECK_DELEGATE_TASK_CAPACITY
```
xingchi-jin, line 68: as we've discussed. i don't think we need to change the ui downloaded yaml files.

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49060)* status: 8 file(s) to review. 1 change requests. 10. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 318 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227364139:-NXqGOW4CEnjzNdUJf4a:b-nea48b) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L318)):*

use boolean for "checkDelegateCapacity"
</blockquote></details>

Going to use just one ENV var that gives the delegate task capacity.
___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 1168 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227413944:-NXqGMht-j-P6OK7C3AP:b-m3xt3s) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L1168)):*

+1
</blockquote></details>

Okay got it.
___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 2052 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227476762:-NXpiDSs3b43d1YPE9d-:bh4vvec) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L2052)):*

Did we not decide to use currentlyExecutingTasks map?
</blockquote></details>

Yeah I have updated this logic now. We need more handling and taken into account the current tasks being acquired as well.
___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 2052 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227365300:-NXpiJdP8tZgFxxrrrhg:btlxpdn) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L2052)):*

log.info
</blockquote></details>

Will add it.
___
*[`400-rest/src/main/java/io/harness/iterator/FailDelegateTaskIteratorHelper.java` line 96 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227363523:-NXpjGlCANG8_6tMM3R9:b-3ljj1) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/400-rest/src/main/java/io/harness/iterator/FailDelegateTaskIteratorHelper.java#L96)):*

Don’t understand this change 
</blockquote></details>

This was a bug that was found while testing. While marking the task as failed here we need to check if the broadcast round has exceeded the max broadcast count instead of checking if it is less.
___
*[`400-rest/src/main/java/software/wings/app/DelegateQueueTask.java` line 159 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227478316:-NXpjhCwBivklB9SWA5R:b-xt0yte) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/400-rest/src/main/java/software/wings/app/DelegateQueueTask.java#L159)):*

Don't we want an upper bound on this?
</blockquote></details>

We have task expiry time which has an upper bound. 
___
*[`400-rest/src/main/resources/delegatetemplates/delegate.sh.ftl` line 112 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227376381:-NXqDnJs8WqmgO95QFIN:b-31orbt) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/400-rest/src/main/resources/delegatetemplates/delegate.sh.ftl#L112)):*

@Raghav-Murali  i think @raghuAtWings 's point here is, if the config is directly read from ENV using the java code, instead of using config, then the change can easily apply to both imm and legacy delegate. 
I think it's a good approach to take. 
</blockquote></details>

Got it. So I will remove it from config then and then read directly from ENV var. We'll ask the customer to setup this ENV var in their delegates.
___
*[`400-rest/src/main/resources/delegatetemplates/common/delegate-environment.ftl` line 68 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227497134:-NXqGUpP3MLpi28Mm6wo:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/400-rest/src/main/resources/delegatetemplates/common/delegate-environment.ftl#L68)):*

as we've discussed. i don't think we need to change the ui downloaded yaml files. 
</blockquote></details>

Done

gn-harness (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49060)* status: 8 file(s) to review. 2 change requests. 9. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 318 at r4](https://reviewable.io/reviews/harness/harness-core/49060#-NXrkxvHES0gok6vX0w-:-NXrkxvHES0gok6vX0w0:bzblw5w) ([raw file](https://github.com/harness/harness-core/blob/711511fee3f571c038a6b362e348a4b3cd8d5885/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L318)):*
> ```Java
>   private final boolean dynamicRequestHandling = isNotBlank(System.getenv().get("DYNAMIC_REQUEST_HANDLING"))
>       && Boolean.parseBoolean(System.getenv().get("DYNAMIC_REQUEST_HANDLING"));
>   private final int delegateTaskCapacity = getDelegateTaskCapacity();
> ```

nit: You can chose to store Optional<Integer> - that makes it cleaner, so that you don't have to assume anything about 0 value



_Code quote:_
```Java
 int
```
___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 2050 at r4](https://reviewable.io/reviews/harness/harness-core/49060#-NXrkXgz2_8QH4fj4Hlk:-NXrkXgz2_8QH4fj4Hll:b-oq64rf) ([raw file](https://github.com/harness/harness-core/blob/711511fee3f571c038a6b362e348a4b3cd8d5885/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L2050)):*
> ```Java
>           // Check if current executing tasks plus currently
>           // acquiring tasks is below capacity.
>           if ((currentlyExecutingTasks.size() + currentlyAcquiringTasks.size()) >= delegateTaskCapacity) {
> ```

Just keep this one and remove the first check.

In the log.info - you can record all 3 values. currently.size, acquiring.size and capacity



_Code quote:_
```Java
 if ((currentlyExecutingTasks.size() + currentlyAcquiringTasks.size()) >= delegateTaskCapacity) {
```
___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 2368 at r4](https://reviewable.io/reviews/harness/harness-core/49060#-NXrktH80T0gc6-Sb497:-NXrktH80T0gc6-Sb498:bw36c93) ([raw file](https://github.com/harness/harness-core/blob/711511fee3f571c038a6b362e348a4b3cd8d5885/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L2368)):*
> ```Java
> 
>   private static int getDelegateTaskCapacity() {
>     // By default, set the task capacity to 10.
> ```

Did you mean 0 or 10?



_Code quote:_
```Java
set the task capacity to 10.
```
___
*[`400-rest/src/main/java/io/harness/iterator/FailDelegateTaskIteratorHelper.java` line 96 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227363523:-NXrlUX41gYWs9exFK6a:bi94ics) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/400-rest/src/main/java/io/harness/iterator/FailDelegateTaskIteratorHelper.java#L96)):*

This was a bug that was found while testing. While marking the task as failed here we need to check if the broadcast round has exceeded the max broadcast count instead of checking if it is less.
</blockquote></details>

I would suggest move this in a separate PR or add tests for it in the same PR. 
___
*[`400-rest/src/main/java/software/wings/app/DelegateQueueTask.java` line 159 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227478316:-NXrlf869O2XtEmp84en:b-82g6go) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/400-rest/src/main/java/software/wings/app/DelegateQueueTask.java#L159)):*

We have task expiry time which has an upper bound. 
</blockquote></details>

nit: Do you have a sense of how long these expiration times can be? I'm not blocking you on this one, but we need to understand whether those expiration times are too low or too high. If too low, no queuing will happen - If too high, then we will have an indeterministic time duration for different types of tasks.  Also, task expiration time could just be the time to execute - if a task gets queued for an hour and then may not get enough time to execute it. Expiration time should ideally start after a task is assigned.

260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java
```diff
@@ -2034,6 +2035,24 @@ private void dispatchDelegateTask(DelegateTaskEvent delegateTaskEvent) {
           return;
         }
 
+        // delegateTaskCapacity equal to 0 will be treated as infinity.
+        // So this feature works only if ENV - DELEGATE_TASK_CAPACITY is defined.
+        if (delegateTaskCapacity != 0) {
+          // We're using the concurrent hashmap size to determine as its thread safe.
+          // Check if current executing tasks is below capacity.
+          if (currentlyExecutingTasks.size() >= delegateTaskCapacity) {
```
xingchi-jin, line 2043: why not remove this if block and just keep the one below ? for logging purpose ?

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49060)* status: 8 file(s) to review. 2 change requests. 14. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 318 at r4](https://reviewable.io/reviews/harness/harness-core/49060#-NXrkxvHES0gok6vX0w-:-NXwluVN1nfe51Fc-PZG:bju0oc8) ([raw file](https://github.com/harness/harness-core/blob/711511fee3f571c038a6b362e348a4b3cd8d5885/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L318)):*

nit: You can chose to store Optional<Integer> - that makes it cleaner, so that you don't have to assume anything about 0 value
</blockquote></details>

That is a good suggestion, thanks! Will update it.
___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 2043 at r4](https://reviewable.io/reviews/harness/harness-core/49060#gh-1228942866:-NXwm7Cr3ibHzsnv5XN-:b-jee7q6) ([raw file](https://github.com/harness/harness-core/blob/711511fee3f571c038a6b362e348a4b3cd8d5885/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L2043)):*

why not remove this if block and just keep the one below ? for logging purpose ?
</blockquote></details>

No, I added for optimization :). So that we need not read the acquiring task count and do an add operation all the time. If you feel it is not needed then I can remove it.
___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 2050 at r4](https://reviewable.io/reviews/harness/harness-core/49060#-NXrkXgz2_8QH4fj4Hlk:-NXwnIgLA0f0m7-1bgBM:bb6un1j) ([raw file](https://github.com/harness/harness-core/blob/711511fee3f571c038a6b362e348a4b3cd8d5885/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L2050)):*

Just keep this one and remove the first check.

In the log.info - you can record all 3 values. currently.size, acquiring.size and capacity
</blockquote></details>

Done
___
*[`260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java` line 2368 at r4](https://reviewable.io/reviews/harness/harness-core/49060#-NXrktH80T0gc6-Sb497:-NXwnOpB34dRTyXiajvn:bwvi3dq) ([raw file](https://github.com/harness/harness-core/blob/711511fee3f571c038a6b362e348a4b3cd8d5885/260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java#L2368)):*

Did you mean 0 or 10?
</blockquote></details>

Removed the comment since with Optional we need not assume any default value.
___
*[`400-rest/src/main/java/io/harness/iterator/FailDelegateTaskIteratorHelper.java` line 96 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227363523:-NXwn_5V7qcKHE9h0Wn1:b-unvc1m) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/400-rest/src/main/java/io/harness/iterator/FailDelegateTaskIteratorHelper.java#L96)):*

I would suggest move this in a separate PR or add tests for it in the same PR. 
</blockquote></details>

For now, I will move it to separate PR as it will speed up this PR.
___
*[`400-rest/src/main/java/software/wings/app/DelegateQueueTask.java` line 159 at r3](https://reviewable.io/reviews/harness/harness-core/49060#gh-1227478316:-NXwnvxvA9pLH8naC4NM:bzddznu) ([raw file](https://github.com/harness/harness-core/blob/6d565cdd8a0740654a9db21476c4b6edc95d26fb/400-rest/src/main/java/software/wings/app/DelegateQueueTask.java#L159)):*

nit: Do you have a sense of how long these expiration times can be? I'm not blocking you on this one, but we need to understand whether those expiration times are too low or too high. If too low, no queuing will happen - If too high, then we will have an indeterministic time duration for different types of tasks.  Also, task expiration time could just be the time to execute - if a task gets queued for an hour and then may not get enough time to execute it. Expiration time should ideally start after a task is assigned.
</blockquote></details>

Yeah we enforce Task timeout while executing the task. This will ensure that the task is expired if it doesn't end within the task timeout configured. We need to ask the customer what is the task timeout they are configuring for their Terraform and other tasks. But for now if the timeout is set to 30 mins then we will just rebroadcast 30 times and we should indicate this to the customer so they can configure a valid timeout.

260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java
```diff
@@ -2034,7 +2038,31 @@ private void dispatchDelegateTask(DelegateTaskEvent delegateTaskEvent) {
...
+          // Check if current executing tasks plus currently
+          // acquiring tasks is below capacity.
+          final int acquiringTaskCount = currentlyAcquiringTasksCount.get();
+          if ((executingTaskCount + acquiringTaskCount) >= taskCapacity) {
+            log.info(
+                "Not acquiring task - currently executing {} and acquiring {} tasks count exceeds task capacity {}",
+                executingTaskCount, acquiringTaskCount, taskCapacity);
+            return;
+          }
+        }
```
gn-harness, line 2063: not sure why we have two blocks. Just keep the second one and log all 3 values. If acquiring task count is 0 - that would give the same effect

260-delegate/src/main/java/io/harness/delegate/service/DelegateAgentServiceImpl.java
```diff
@@ -2840,4 +2885,26 @@ private void checkForSSLCertVerification(String accountId) {
       log.error("SSL Cert Verification failed with exception ", e);
     }
   }
+
+  private void addIntoCurrentlyAcquiringTasks(String delegateTaskId) {
+    currentlyAcquiringTasks.add(delegateTaskId);
+    updateCounterIfLessThanCurrent(currentlyAcquiringTasksCount, currentlyAcquiringTasks.size());
```
gn-harness, line 2891: this is not the right thing to do - you are using a function which was meant to do something else.

Raghav-Murali:
trigger uts

xingchi-jin (approved):
> Reviewed 3 of 9 files at r2, 3 of 6 files at r4, 3 of 5 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49060)* status: 4 file(s) to review. 2 change requests. 14. unresolved discussions. 1 of 2 approvals obtained.

raghuAtWings (approved):

gn-harness (approved):

Raghav-Murali:
trigger messagecheck

PR #49129, opened 2023-06-12
fix: [CI-1382]: update libraries

rafaelmart1ns:
## Describe your changes
bump the golang version to 1.19, redis to v9 and aws-go and add localhost in local config to prevent popup window when ran a server

## Checklist

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/49129)


product/log-service/go.mod
```diff
@@ -1,42 +1,80 @@
 module github.com/harness/harness-core/product/log-service
 
-go 1.14
+go 1.20
```
lasdox, line 3: i suggest we set go.mod as 1.19 to not force developers to update to 1.20 right away

rafaelmart1ns: it was changed.

rafaelmart1ns:
trigger gobuild

rafaelmart1ns:
trigger gobuild

rafaelmart1ns:
trigger gobuild

rafaelmart1ns:
trigger gobuild

rafaelmart1ns:
trigger uts

rafaelmart1ns:
trigger uts

product/log-service/stream/redis/redis.go
```diff
@@ -120,28 +119,28 @@ func (r *Redis) Create(ctx context.Context, key string) error {
 	// macro node. MaxLen will always be >= 5000 but can be a few tens of entries
 	// more as well.
 	args := &redis.XAddArgs{
-		Stream:       key,
-		ID:           "*",
-		MaxLenApprox: maxStreamSize,
-		Values:       map[string]interface{}{entryKey: []byte{}},
+		Stream: key,
+		ID:     "*",
+		MaxLen: maxStreamSize,
```
vistaarjuneja, line 124: `MaxLenApprox` is more efficient than `MaxLen`. We don't need to use `MaxLen` here.
```However trimming with MAXLEN can be expensive: streams are represented by macro nodes into a radix tree, in order to be very memory efficient. Altering the single macro node, consisting of a few tens of elements, is not optimal. So it's possible to use the command in the following special form:

XADD mystream MAXLEN ~ 1000 * ... entry fields here ...```
From documentation ^

File: product/log-service/stream/redis/redis.go [hunk 3]
```diff
@@ -160,18 +159,18 @@ func (r *Redis) Write(ctx context.Context, key string, lines ...*stream.Line) er
 	for _, line := range lines {
 		bytes, _ := json.Marshal(line)
 		arg := &redis.XAddArgs{
-			Stream:       key,
-			Values:       map[string]interface{}{entryKey: bytes},
-			MaxLenApprox: maxStreamSize,
-			ID:           "*",
+			Stream: key,
+			Values: map[string]interface{}{entryKey: bytes},
+			MaxLen: maxStreamSize,
```
vistaarjuneja, line 164: Ditto

product/log-service/stream/redis/redis.go
```diff
@@ -385,7 +384,7 @@ func (r *Redis) setExpiry(key string, expiry time.Duration) error {
 }
 
 // Scan all the keys and set an expiry on them if it's not set
-func (r *Redis) expiryWatcher(expiry time.Duration) {
+func (r *Redis) expiryWatcher(ctx context.Context, expiry time.Duration) {
```
vistaarjuneja, line 387: We need to pass ctx to the function which calls `expiryWatcher` as well - can't see that change?

product/log-service/config/config.go
```diff
@@ -25,7 +25,7 @@ type Config struct {
 	}
 
 	Server struct {
-		Bind  string `envconfig:"LOG_SERVICE_HTTP_BIND" default:":8079"`
+		Bind  string `envconfig:"LOG_SERVICE_HTTP_BIND" default:"localhost:8079"`
```
vistaarjuneja, line 28: why is this required and what is the effect?

product/log-service/store/s3/s3.go
```diff
@@ -84,49 +106,52 @@ func (s *Store) Download(ctx context.Context, key string) (io.ReadCloser, error)
...
 		u.PartSize = 32 * 1024 * 1024 // 32MB per part
 	})
 	keyWithPrefix := path.Join("/", s.prefix, key)
-	input := &s3manager.UploadInput{
-		ACL:    aws.String(s.acl),
+	input := &s3.PutObjectInput{
 		Bucket: aws.String(s.bucket),
 		Key:    aws.String(keyWithPrefix),
 		Body:   r,
+		ACL:    types.ObjectCannedACL(s.acl),
```
vistaarjuneja, line 131: please double check this parameter for `acl` here and why the type has been changed in v2 - and what this means for us (is there anything backward incompatible here). what does `ObjectCannedACL` expect as input

product/log-service/store/s3/s3.go
```diff
@@ -84,49 +106,52 @@ func (s *Store) Download(ctx context.Context, key string) (io.ReadCloser, error)
...
-	req, _ := svc.PutObjectRequest(&s3.PutObjectInput{
+	req, _ := s.presignService.PresignPutObject(ctx, &s3.PutObjectInput{
 		Bucket: aws.String(s.bucket),
 		Key:    aws.String(keyWithPrefix),
+	}, func(opts *s3.PresignOptions) {
+		opts.Expires = time.Duration(int64(expire.Seconds()) * int64(time.Second))
 	})
-	return req.Presign(expire)
+
+	return req.URL, nil
```
vistaarjuneja, line 148: we should not always return `nil` - return error if it is not nil

product/log-service/store/s3/s3.go
```diff
@@ -9,69 +9,91 @@ package s3
...
+	presignService *s3.PresignClient
 }
 
 // NewEnv returns a new S3 log store from the environment.
 func NewEnv(bucket, prefix, endpoint string, pathStyle bool, accessKeyID, accessSecretKey, region, acl string) *Store {
-	disableSSL := false
-
 	if endpoint != "" {
-		disableSSL = !strings.HasPrefix(endpoint, "https://")
+		http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```
vistaarjuneja, line 44: why are we setting `InsecureSkipVerify` as true? This will be susceptible to MITM attacks

product/log-service/store/s3/s3.go
```diff
@@ -84,49 +106,52 @@ func (s *Store) Download(ctx context.Context, key string) (io.ReadCloser, error)
...
+
+	return req.URL, nil
 }
 
 // Delete purges the log stream from the S3 datastore.
 func (s *Store) Delete(ctx context.Context, key string) error {
-	svc := s3.New(s.session)
 	keyWithPrefix := path.Join("/", s.prefix, key)
-	_, err := svc.DeleteObject(&s3.DeleteObjectInput{
+	_, err := s.service.DeleteObject(ctx, &s3.DeleteObjectInput{
```
vistaarjuneja, line 154: seems like we used to generate a new session for each call but now we are using the same service. can we check that this is concurrency safe with different keys?

product/log-service/stream/redis/redis.go
vistaarjuneja: https://redis.io/docs/data-types/streams-tutorial/

product/log-service/store/s3/s3.go
```diff
@@ -84,49 +106,52 @@ func (s *Store) Download(ctx context.Context, key string) (io.ReadCloser, error)
 // DownloadLink creates a pre-signed link that can be used to
 // download the logs to the S3 datastore.
 func (s *Store) DownloadLink(ctx context.Context, key string, expire time.Duration) (string, error) {
-	svc := s3.New(s.session)
 	keyWithPrefix := path.Join("/", s.prefix, key)
-	req, _ := svc.GetObjectRequest(&s3.GetObjectInput{
+	req, _ := s.presignService.PresignGetObject(ctx, &s3.GetObjectInput{
```
lasdox, line 110: instead of hidding the error ,lets return it if err != nil

product/log-service/stream/redis/redis.go
lasdox: yes, @rafaelmart1ns need to change on the scheduler method line 105

product/log-service/store/s3/s3.go
lasdox: left over probably

product/log-service/store/s3/s3.go
lasdox: No backward changes, I dont think we should use the constructor here, AWS has predefined the values which we can use, so we should parse and use the constants here.
One thing I noticed is that log-service right now is not using this property, so we should be good for now

product/log-service/store/s3/s3.go
lasdox: This is not true @vistaarjuneja, old code just created one session for all requests
Sessions have nothing to do with Concurrency 
They are used for scoped configs, since we are not changing configuration, e.g: region, cluster
we should only use one Configuration (similar to session)

product/log-service/config/config.go
rafaelmart1ns: it just prevents the popup window to gave on macos system.

product/log-service/store/s3/s3.go
rafaelmart1ns: It was updated.

product/log-service/store/s3/s3.go
rafaelmart1ns: it was updated.

product/log-service/store/s3/s3.go
rafaelmart1ns: it was updated.

product/log-service/stream/redis/redis.go
rafaelmart1ns: Yes, you are right. It was solved by adding the `approx` as true.

product/log-service/stream/redis/redis.go
rafaelmart1ns: ditto

product/log-service/stream/redis/redis.go
vistaarjuneja: @rafaelmart1ns don't see that change yet? this is invoked on L105. Can you confirm after that change that this thread is running?

product/log-service/store/s3/s3.go
vistaarjuneja: Don't see the updated change

product/log-service/store/s3/s3.go
```diff
@@ -84,49 +107,58 @@ func (s *Store) Download(ctx context.Context, key string) (io.ReadCloser, error)
...
 // upload the logs to the S3 datastore.
 func (s *Store) UploadLink(ctx context.Context, key string, expire time.Duration) (string, error) {
-	svc := s3.New(s.session)
 	keyWithPrefix := path.Join("/", s.prefix, key)
-	req, _ := svc.PutObjectRequest(&s3.PutObjectInput{
+	req, err := s.presignService.PresignPutObject(ctx, &s3.PutObjectInput{
 		Bucket: aws.String(s.bucket),
 		Key:    aws.String(keyWithPrefix),
+	}, func(opts *s3.PresignOptions) {
+		opts.Expires = time.Duration(int64(expire.Seconds()) * int64(time.Second))
```
vistaarjuneja, line 149: `expire` is of `time.Duration` so you don't need this conversion

product/log-service/store/s3/s3.go
vistaarjuneja: @rafaelmart1ns can you confirm on the above once? Do double check whether we have acl set in `ng-prod-manifests` anywhere and whether we need the constructor

product/log-service/store/s3/s3.go
vistaarjuneja: @lasdox you are right - we created one session but looks like service was created each time. But now we use a single service. 
This is what I was looking for: https://aws.github.io/aws-sdk-go-v2/docs/making-requests/#concurrently-using-service-clients and looks like this is ok. Can you guys also go through this once cc: @rafaelmart1ns

PR #49503, opened 2023-06-21
feat: [PL-39561]: api def

xingchi-jin:
## Describe your changes
API definition for bijou delegate execute, init 

## Checklist

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/49503)


920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
```diff
@@ -0,0 +1,64 @@
...
+  repeated K8sTaskContainer tasks = 1;  // Container specs for task containers
+  Resource resource = 2;     // Global resource requirements for infra types that need it (e.g. K8S CPU/Mem)
+  map<string, string> env = 3;           // Any global env variables. These envs will be applied to all containers
+  SecurityContext security_context = 4;  // Startup user id for the execution infra
+  string working_dir = 5;                // Working dir that all tasks have read/write access to
+  repeated Resource resources = 6;       // Any global resources like shared volumes
+}
+
+message K8sTaskContainer {
+  string id = 1;           // Id for the step
```
gn-harness, line 34: Why do we need this Id?

gn-harness: And this ID is supposed to be unique? if Yes, unique within K8Infra spec?

920-delegate-service-beans/src/main/proto/io/harness/delegate/resource.proto
```diff
@@ -0,0 +1,65 @@
...
+  RES_FILE = 2;
+}
+
+message EmptyDirVolume {
+  string name = 1;
+  string path = 2;
+  string size = 3;
+  string medium = 4;
+}
+
```
gn-harness, line 45: where are these used?

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
```diff
@@ -0,0 +1,64 @@
...
+  ContainerSpec container_spec = 2;      // Spec for the container environment
+  SecurityContext security_context = 3;  // Startup user id for the execution infra
+}
+
+message ContainerSpec {
+  string image = 1;                   // Docker image tag
+  repeated int64 port = 2;            // Port exposed for the container
```
gn-harness, line 41: While starting up K8 pod, are we summing up the resource requirements from containers? or do we just pick the maximum?

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
```diff
@@ -0,0 +1,64 @@
...
+  ComputingResource resource = 3;     // Resource requirements for infra types that need it (e.g. K8S CPU/Mem). If it's not provided, a default setting will be provided by platform
+  map<string, string> env = 4;        // Env variables
+  repeated string command = 5;        // Commands to be executed
```
gn-harness, line 44: these runs to start the execution?

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
```diff
@@ -0,0 +1,64 @@
...
+  repeated string arg = 6;            // Args to be passed
```
gn-harness, line 45: args to what? to the commands?

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
```diff
@@ -0,0 +1,64 @@
...
+    // VMInfraSpec
+  }
+}
+
+message K8sInfraSpec {
+  repeated K8sTaskContainer tasks = 1;  // Container specs for task containers
+  Resource resource = 2;     // Global resource requirements for infra types that need it (e.g. K8S CPU/Mem)
+  map<string, string> env = 3;           // Any global env variables. These envs will be applied to all containers
+  SecurityContext security_context = 4;  // Startup user id for the execution infra
+  string working_dir = 5;                // Working dir that all tasks have read/write access to
```
xingchi-jin, line 29: maybe we don't need to expose it. 
set default working directory internally

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
```diff
@@ -0,0 +1,64 @@
...
+  LogConfig log_config = 1;   // log stream service settings
+  oneof infraSpec {
+    K8sInfraSpec k8Infraspec = 2;
+    // VMInfraSpec
+  }
+}
+
+message K8sInfraSpec {
+  repeated K8sTaskContainer tasks = 1;  // Container specs for task containers
+  Resource resource = 2;     // Global resource requirements for infra types that need it (e.g. K8S CPU/Mem)
```
xingchi-jin, line 26: remove global compute resource setting

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
```diff
@@ -0,0 +1,64 @@
...
+}
+
+message ContainerSpec {
+  string image = 1;                   // Docker image tag
+  repeated int64 port = 2;            // Port exposed for the container
+  ComputingResource resource = 3;     // Resource requirements for infra types that need it (e.g. K8S CPU/Mem). If it's not provided, a default setting will be provided by platform
+  map<string, string> env = 4;        // Env variables
+  repeated string command = 5;        // Commands to be executed
+  repeated string arg = 6;            // Args to be passed
+  string working_dir = 7;             // Working dir that all tasks have read/write access to
```
xingchi-jin, line 46: maybe remove it

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
xingchi-jin: i removed it for now. if ci runner specifically need step id, it can put in env

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
xingchi-jin: i double checked k8s document, both command and args are array. https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#define-a-command-and-arguments-when-you-create-a-pod 
then i think it makes sense here to use repeated.

Raghav-Murali (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49503)* status: 8 file(s) to review. 2. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/BUILD.bazel` line 22 at r2](https://reviewable.io/reviews/harness/harness-core/49503#-NYV9rte00eqK4iHnraZ:-NYV9rte00eqK4iHnra_:b-kj53zn) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/BUILD.bazel#L22)):*
> ```Starlark
>         "log_config.proto",
>         "resource.proto",
>         "secrets.proto"
> ```

nit: Won't *.proto work here ? That way we need not keep updating this for every new file.
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto` line 20 at r2](https://reviewable.io/reviews/harness/harness-core/49503#-NYVB8bT1WZyogp-0kGn:-NYVB8bT1WZyogp-0kGo:bywg6dj) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto#L20)):*
> ```proto
>     // VMExecute
>   }
>   string infraRef = 2; // execution infra id created by calling SetupExecutionInfrastructureRequest
> ```

string infraRefId ?
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto` line 23 at r2](https://reviewable.io/reviews/harness/harness-core/49503#-NYVAqFpETeLW_7pQETi:-NYVAqFpETeLW_7pQETj:bbeeikt) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto#L23)):*
> ```proto
> }
> message K8sExecute {
>   RunConfig taskData = 1;      // task binary data
> ```

nit: Can we just have bytes taskData for now and maybe extend it to another message in future if more fields will be added ?
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/secrets.proto` line 15 at r2](https://reviewable.io/reviews/harness/harness-core/49503#-NYVCI-a-EWvjN_3Aj60:-NYVCI-a-EWvjN_3Aj61:b-tduu2i) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/secrets.proto#L15)):*
> ```proto
>   string address = 1;
>   string username = 2;
>   string password = 3;
> ```

So we will be sending the decrypted secrets from CI / Pipeline service to the manager / DMS ? If we're going to decrypt the secret in DMS or delegate then we can have RegistryAuth as bytes right ?

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
```diff
@@ -0,0 +1,61 @@
...
+  repeated string command = 5;        // Commands to be executed
+  repeated string args = 6;            // Args to be passed
+}
+
+/*
+  Defines security context for running a task. Note these values would always be there (unless entire context is
+  omitted) with their defaults. E.g.
+    privileged = false; runAsUser = 0 (root); runAsNonRoot = false; allowPrivilegeEscalation = false;
+ */
+message SecurityContext {
```
xingchi-jin, line 51: move to separate file, and check each field for defaults

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
xingchi-jin: will change comment to "command"

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
```diff
@@ -0,0 +1,61 @@
...
+  oneof infraSpec {
+    K8sInfraSpec k8Infraspec = 2;
+    // VMInfraSpec
+  }
+}
+
+message K8sInfraSpec {
+  repeated K8sTaskContainer tasks = 1;  // Container specs for task containers
+  map<string, string> env = 3;           // Any global env variables. These envs will be applied to all containers
+  SecurityContext security_context = 4;  // Startup user id for the execution infra
```
xingchi-jin, line 27: remove for now until figure out what is security context

920-delegate-service-beans/src/main/proto/io/harness/delegate/resource.proto
xingchi-jin: discussed in slack. resolve

gn-harness (approved):

xingchi-jin (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49503)* status: 9 file(s) to review. 9. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/BUILD.bazel` line 22 at r2](https://reviewable.io/reviews/harness/harness-core/49503#-NYV9rte00eqK4iHnraZ:-NYVaxbK3mDdCtJohxkC:b-wd36b0) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/BUILD.bazel#L22)):*

nit: Won't *.proto work here ? That way we need not keep updating this for every new file.
</blockquote></details>

i feel it's better to be explicit to avoid any random mistakes. but it's fine both 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto` line 20 at r2](https://reviewable.io/reviews/harness/harness-core/49503#-NYVB8bT1WZyogp-0kGn:-NYVb3hCCoGXMaKcoe3k:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto#L20)):*

string infraRefId ?
</blockquote></details>

Done.
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto` line 24 at r2](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237223818:-NYVb5Hs68sVo0RNvy71:b-896fix) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto#L24)):*

Please document the behavior when both infraRef and infraSpec are specified
</blockquote></details>

Done.
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/secrets.proto` line 15 at r2](https://reviewable.io/reviews/harness/harness-core/49503#-NYVCI-a-EWvjN_3Aj60:-NYVbbfc635pGdXD2P3L:bbjgjyq) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/secrets.proto#L15)):*

So we will be sending the decrypted secrets from CI / Pipeline service to the manager / DMS ? If we're going to decrypt the secret in DMS or delegate then we can have RegistryAuth as bytes right ?
</blockquote></details>

we will define separate secret section in API. DMS and delegate will decrypt them before pass thru env to runners. I added a TODO here
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 41 at r2](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237306291:-NYVbt6A1EnPgQ59YZFe:b-6to8u5) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L41)):*

While starting up K8 pod, are we summing up the resource requirements from containers? or do we just pick the maximum?
</blockquote></details>

resource is per container. I don't see k8s runner summing up that.

xingchi-jin:
trigger uts

Raghav-Murali (dismissed):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49503)* status: 9 file(s) to review. 9. unresolved discussions. 1 of 2 approvals obtained.

mdmilic (commented):
> Reviewed 3 of 8 files at r1, 1 of 3 files at r3, 3 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49503)* status: 9 file(s) to review. 9. unresolved discussions. 1 of 2 approvals obtained. 

___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto` line 16 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYFZzA8CnjFAwRDXy3:-NYYFZzA8CnjFAwRDXy4:b-o9m8oe) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto#L16)):*
> ```proto
>   oneof executionParams {
>     K8sExecute k8sExecute = 1;
>     // VMExecute
> ```

would there be any difference between K8S, VM and other execute related to task input? I think task input should be independent of the infra where the task is executing, so this can just be bytes 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/log_config.proto` line 14 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYIWpc9hjqtOGOroWC:-NYYIWpc9hjqtOGOroWD:b-yhb8gu) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/log_config.proto#L14)):*
> ```proto
> message LogConfig {
>   // log service url
>   string url = 1;
> ```

I don't think module should provide url, this is delegate/runner startup config. 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/log_config.proto` line 18 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYItq50etxJQgxa27-:-NYYItq50etxJQgxa270:b-9yg33u) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/log_config.proto#L18)):*
> ```proto
>   string log_key = 2;
>   // log service token
>   string token = 3;
> ```

Token seems like something DMS should be able to provide on its own (i.e. is it unrelated to task, task input or anything module specific)? We could use this in initial implementation though if easier. Or is it being provided by pipeline service (and not module service).
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/pod_security.proto` line 18 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYYT08Eu6O9C6JeSN8:-NYYYT08Eu6O9C6JeSN9:b3xd7lb) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/pod_security.proto#L18)):*
> ```proto
>     privileged = false; runAsUser = 0 (root); runAsNonRoot = false;
>  */
> message SecurityContext {
> ```

These are all the fields that CI UI supports for the security context, we need to add all of them. 
```  
  bool allowPrivilegeEscalation = 1;
  bool privileged = 2;
  string procMount = 3;
  bool readOnlyRootFilesystem = 4;
  bool runAsNonRoot = 5;
  int64 runAsGroup = 6;
  int64 runAsUser = 7;
  repeated string add_capability = 8;
  repeated string drop_capability = 9;
```
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/secrets.proto` line 19 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYHss279zRoNmQNJDa:-NYYHss279zRoNmQNJDb:b2fj45t) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/secrets.proto#L19)):*
> ```proto
> }
> 
> message ImagePull {
> ```

I don't think this would belong to Secret.proto. Also, I don't think we should bundle image and auth like this. lastly, both of these doesn't seem used anywhere :smile: 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 33 at r1](https://reviewable.io/reviews/harness/harness-core/49503#-NYYhQAn57RiJ6rF_EIg:-NYYhQAn57RiJ6rF_EIh:banoqn3) ([raw file](https://github.com/harness/harness-core/blob/c5f01061736c2eb781534368fabea36253032413/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L33)):*
> ```proto
> }
> 
> message K8sTaskContainer {
> ```

You still need:
```
  ExecutionMode mode = 2;  // Is step going to be executed once or periodically
  ExecutionPriority priority =
      3;  // Priority of the step. Under heavier load, higher priority steps may have precedence
  ExecutionType execution_type = 4;  // Type of the step
  StepRuntime runtime = 5;           // Environment in which the task should run
  repeated Secret input_secrets =
```
`mode` & `priority` are not used yet but are designed to support sync and perpetual task implementation. You can remove them if you want. 
`execution_type` - is needed now (though not yet implemented). CI supports background tasks & plugins as a container types. background tasks are for e.g. test mongoDB and are always running on the Pod. plugin types are one off, running to completion (the ones we implemented so far). 
`input_secrets` are needed as they are task secrets. 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 26 at r2](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237643811:-NYYMKUnAoI5cWh6JuW_:b-k6yhj0) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L26)):*

remove global compute resource setting 
</blockquote></details>

I need global compute resources to be able to give them to the LE as Pod resources. DMS and Delegate/Runner are not capable of computing the stage execution plan and not capable of calculating how much resources the Pod needs (i.e. the logic where CI does something like `max(sequential task, sum(parallel tasks) )`). I currently still rely on the module (e.g. CI/CD) calculating this (or probably pipeline service itself creates the plan & CI just applies their calculation), and then providing it as global resource. We could look into doing this from DMS with some Pipeline APIs maybe in the future. 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 29 at r2](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237643468:-NYYNnd65qw0FJATuDxA:b-yzgym7) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L29)):*

maybe we don't need to expose it. 
set default working directory internally 
</blockquote></details>

CI supports setting a custom path as a working dir. If not provided, runner would setup the default path (`/harness/`) as workdir, but this is to allow user to specify it. However, I'm not finding this UI option now, so not sure if specific to some special use cases or it got removed. Let's remove it for now. 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 34 at r2](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237301834:-NYYOSYL3LcbzPHvaaTB:b-41qplf) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L34)):*

i removed it for now. if ci runner specifically need step id, it can put in env 
</blockquote></details>

this `id` is `taskId` from the current world and it is needed because it becomes the name for the container (and is used as a base for a lot of other naming and operations like task routing). 
This taskId is generated though by DMS (AFAIK) when the task is scheduled, so wouldn't make sense for it to be here, but still DMS would have to inject it somehow before forwarding data to the runner. 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 41 at r2](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237306291:-NYY__xH1a64fOq9SG1J:blpdhhn) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L41)):*

resource is per container. I don't see k8s runner summing up that. 
</blockquote></details>

the sum would come in as pod level compute resource form the manager. 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 44 at r2](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237648697:-NYYe7j158aRpcnrnJ6Z:b-2lltf8) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L44)):*

command can be just a string, because it's just one command 
</blockquote></details>

It is possible for it to be array (e.g. `["/bin/bash", "-c", "echo hello"]') and K8S API expects it as an array
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 27 at r3](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237815323:-NYYO5JK5Lo8aPnQ24EW:b-6icspr) ([raw file](https://github.com/harness/harness-core/blob/1a623fc365b05b10af7eb9e86a09657c5c8b9ceb/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L27)):*

remove for now until figure out what is security context 
</blockquote></details>

CI UI supports setting global (i.e. Pod) security context. It also supports setting Step level (i.e. container) security context, so this is needed. 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 25 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYMFHNBHfO9ZafxfjU:-NYYMFHNBHfO9ZafxfjV:b-ozdkf7) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L25)):*
> ```proto
> }
> 
> message K8sInfraSpec {
> ```

K8S Spec should go into a separate proto file I think (VM and all others we define also separate files). Also, why changing the K8SSpec so much, I thought you would just move/rename the infra.proto from 955 to 920. 
While some renaming and refactoring is fine, all the fields that were there initially were required based on the current CI featureset and the runner implementation, the only exception is taskId mentioned below which was needed for `DMS -> Delegate` communication (i.e. in 955), but not for `Module -> DMS` (i.e. here)
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 25 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYjaq9-dAoSHXLw1Mv:-NYYjaqA76YzptNneEZW:b3xxjkz) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L25)):*
> ```proto
> }
> 
> message K8sInfraSpec {
> ```

You are missing timeout here. 
I know we added it to the SchedulingConfig, but that actually doesn't work because you need timeout per task and not for entire stage (task group).
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 32 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYQoaL-CNxGBtd725T:-NYYQoaL-CNxGBtd725U:b-4v9ohx) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L32)):*
> ```proto
> 
> message K8sTaskContainer {
>   ContainerSpec container_spec = 1;      // Spec for the container environment
> ```

rename variable to just `spec`
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 33 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYQkKL1Hwm6Gm5hpc3:-NYYQkKL1Hwm6Gm5hpc4:b-ntxgtj) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L33)):*
> ```proto
> message K8sTaskContainer {
>   ContainerSpec container_spec = 1;      // Spec for the container environment
>   SecurityContext security_context = 2;  // Startup user id for the execution infra
> ```

Why not adding SecurityContext inside spec
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 38 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYY_gj-ETYd4GyqRaNH:-NYY_gj-ETYd4GyqRaNI:b-j91iw0) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L38)):*
> ```proto
> message ContainerSpec {
>   string image = 1;                   // Docker image tag
>   repeated int64 port = 2;            // Port exposed for the container
> ```

what's the port use case? I didn't find a way to configure any ports for steps. The only ones I saw are the ones which are done by lite runner now so not needed here.

mdmilic (commented):
if possible try to add some extra packages, we will end up with a lot of files in a single package which is not good for deps. 
E.g. infra, infra.k8s, infra.vm, exec etc...

> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49503#-:-NYYnSZx893C-2hu0AMv:bxidkbf)* status: 9 file(s) to review. 9. unresolved discussions. 1 of 2 approvals obtained. 

___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/pod_security.proto` line 18 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYnAAa2khrYJZAjjRV:-NYYnAAa2khrYJZAjjRW:b-yp8iwl) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/pod_security.proto#L18)):*
> ```proto
>     privileged = false; runAsUser = 0 (root); runAsNonRoot = false;
>  */
> message SecurityContext {
> ```

I don't think this file should be called pod_security as it's not pod specific. You can set it on container as well, but probably bunch of things are just Linux specific (i.e. also same on VM), not sure if all though.

xingchi-jin:
trigger uts

xingchi-jin:
trigger uts

xingchi-jin (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49503)* status: 9 file(s) to review. 20. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto` line 16 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYFZzA8CnjFAwRDXy3:-NYd0oWgE3qGhEPn_x_1:b-jy1gzn) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/execute.proto#L16)):*

would there be any difference between K8S, VM and other execute related to task input? I think task input should be independent of the infra where the task is executing, so this can just be bytes 
</blockquote></details>

image, volume network are set at execution phase for vms 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/log_config.proto` line 14 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYIWpc9hjqtOGOroWC:-NYd1BE05g0v6WwQdI8I:b-aj28so) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/log_config.proto#L14)):*

I don't think module should provide url, this is delegate/runner startup config. 
</blockquote></details>

for vm / docker runner, ci has a feature if url is provided, it allows log to send to different place instead of log stream service.
but maybe it's a minor use case, agree we can start with default url 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/log_config.proto` line 18 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYItq50etxJQgxa27-:-NYd1UueBRQSp8ijwGkD:b-v0i0zv) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/log_config.proto#L18)):*

Token seems like something DMS should be able to provide on its own (i.e. is it unrelated to task, task input or anything module specific)? We could use this in initial implementation though if easier. Or is it being provided by pipeline service (and not module service).
</blockquote></details>

currently both options are there. eg ci manager fetches token, while some cd tasks rely on manager to provide token. i think let's keep the options open 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/pod_security.proto` line 18 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYYT08Eu6O9C6JeSN8:-NYd1hpaCjVl-l4Tlxif:bo29326) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/pod_security.proto#L18)):*

These are all the fields that CI UI supports for the security context, we need to add all of them. 
```  
  bool allowPrivilegeEscalation = 1;
  bool privileged = 2;
  string procMount = 3;
  bool readOnlyRootFilesystem = 4;
  bool runAsNonRoot = 5;
  int64 runAsGroup = 6;
  int64 runAsUser = 7;
  repeated string add_capability = 8;
  repeated string drop_capability = 9;
```
</blockquote></details>

i see them all set on stage level but don't see them set at step level.
should we create two security context objects in API ? 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/pod_security.proto` line 18 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYnAAa2khrYJZAjjRV:-NYd1sHtBr3Jl4Z6zHTT:b-918szl) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/pod_security.proto#L18)):*

I don't think this file should be called pod_security as it's not pod specific. You can set it on container as well, but probably bunch of things are just Linux specific (i.e. also same on VM), not sure if all though. 
</blockquote></details>

sure, will just rename to security_context.proto 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/secrets.proto` line 19 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYYHss279zRoNmQNJDa:-NYd27on83Xa_vtkKJbH:b8ftibq) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/secrets.proto#L19)):*

I don't think this would belong to Secret.proto. Also, I don't think we should bundle image and auth like this. lastly, both of these doesn't seem used anywhere :smile: 
</blockquote></details>

agree, my secret design is on the way. that's why i put a todo here 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 33 at r1](https://reviewable.io/reviews/harness/harness-core/49503#-NYYhQAn57RiJ6rF_EIg:-NYd438T1ALH-ct4DdtO:b3b2qlw) ([raw file](https://github.com/harness/harness-core/blob/c5f01061736c2eb781534368fabea36253032413/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L33)):*

You still need:
```
  ExecutionMode mode = 2;  // Is step going to be executed once or periodically
  ExecutionPriority priority =
      3;  // Priority of the step. Under heavier load, higher priority steps may have precedence
  ExecutionType execution_type = 4;  // Type of the step
  StepRuntime runtime = 5;           // Environment in which the task should run
  repeated Secret input_secrets =
```
`mode` & `priority` are not used yet but are designed to support sync and perpetual task implementation. You can remove them if you want. 
`execution_type` - is needed now (though not yet implemented). CI supports background tasks & plugins as a container types. background tasks are for e.g. test mongoDB and are always running on the Pod. plugin types are one off, running to completion (the ones we implemented so far). 
`input_secrets` are needed as they are task secrets. 
</blockquote></details>

the first use case is to run a simple CD task. we can add these fields when we add plugin and background use cases. let's start with minimum and every field needs to be useful 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 29 at r2](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237643468:-NYd3dLXAoerHwGpF2uy:b-vdr2ii) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L29)):*

CI supports setting a custom path as a working dir. If not provided, runner would setup the default path (`/harness/`) as workdir, but this is to allow user to specify it. However, I'm not finding this UI option now, so not sure if specific to some special use cases or it got removed. Let's remove it for now. 
</blockquote></details>

i think if we don't find it, and workdir can be internally handled, let's remove it from here. it can be a field in runner's api, so it's flexible to add in the future 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 34 at r2](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237301834:-NYd3L6w8bNygwDtIccf:b-jyzipq) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L34)):*

this `id` is `taskId` from the current world and it is needed because it becomes the name for the container (and is used as a base for a lot of other naming and operations like task routing). 
This taskId is generated though by DMS (AFAIK) when the task is scheduled, so wouldn't make sense for it to be here, but still DMS would have to inject it somehow before forwarding data to the runner. 
</blockquote></details>

got it, it can be handled internally 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 44 at r2](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237648697:-NYd4t-A-WFMC1egj4W1:b9ksczp) ([raw file](https://github.com/harness/harness-core/blob/bf10a5206e9b124cc007b360f7052d218d7fbc77/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L44)):*

It is possible for it to be array (e.g. `["/bin/bash", "-c", "echo hello"]') and K8S API expects it as an array
</blockquote></details>

i think it's ok 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 27 at r3](https://reviewable.io/reviews/harness/harness-core/49503#gh-1237815323:-NYd3VaTC_G3jxdDgnPL:b-2jsedn) ([raw file](https://github.com/harness/harness-core/blob/1a623fc365b05b10af7eb9e86a09657c5c8b9ceb/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L27)):*

CI UI supports setting global (i.e. Pod) security context. It also supports setting Step level (i.e. container) security context, so this is needed. 
</blockquote></details>

should we create a simpler security context object for step ? 
___
*[`920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto` line 38 at r4](https://reviewable.io/reviews/harness/harness-core/49503#-NYY_gj-ETYd4GyqRaNH:-NYd4kflAWIPPEoJ9fSw:bfsyp17) ([raw file](https://github.com/harness/harness-core/blob/44525aeea0882ffa8b645a30538eb8f7daac32da/920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto#L38)):*

what's the port use case? I didn't find a way to configure any ports for steps. The only ones I saw are the ones which are done by lite runner now so not needed here. 
</blockquote></details>

this is per discussion with @Raghav-Murali , could you confirm ?

xingchi-jin:
trigger uts

xingchi-jin:
trigger ut9

Raghav-Murali (dismissed):

920-delegate-service-beans/src/main/proto/io/harness/delegate/setup_infra.proto
```diff
@@ -0,0 +1,22 @@
...
+
+package io.harness.delegate;
+
+import "920-delegate-service-beans/src/main/proto/io/harness/delegate/k8s_infra.proto";
+import "920-delegate-service-beans/src/main/proto/io/harness/delegate/log_config.proto";
+option go_package = "920_delegate_service_beans";
+
+option java_multiple_files = true;
+
+message ExecutionInfrastructure {
```
gn-harness, line 16: nit: rename to execution_infrastructure.proto

920-delegate-service-beans/src/main/proto/io/harness/delegate/k8s_infra.proto
```diff
@@ -0,0 +1,35 @@
...
+  ComputingResource compute_resource =
+      4;  // Global resource setting for k8s infra. TODO: Marko to figure out resource allocation
+  SecurityContext security_context = 5;  // Optional field to set global security context
+}
+
+message ContainerSpec {
+  string image = 1;                // Docker image tag
+  repeated int64 port = 2;         // Port exposed for the container
+  ComputingResource resource = 3;  // Resource requirements for infra types that need it (e.g. K8S CPU/Mem). If it's not
+                                   // provided, a default setting will be provided by platform
```
gn-harness, line 29: a default setting?

920-delegate-service-beans/src/main/proto/io/harness/delegate/resource.proto
```diff
@@ -0,0 +1,58 @@
...
+  string memory = 1;
+  string cpu = 2;
+}
+
+/*
+  Intended for defining any additional resources the container should use, e.g. additional volume
+ */
+message Resource {
+  google.protobuf.Any spec = 1;
+  ResourceType type = 2;
```
gn-harness, line 27: oneof volume {
  EmptyDirVolume emptyDirVolume = 1;
}

gn-harness (dismissed):

xingchi-jin:
trigger uts

xingchi-jin:
trigger uts

xingchi-jin:
trigger uts

xingchi-jin:
trigger uts

xingchi-jin:
trigger ut2

Raghav-Murali (approved):

PR #49272, opened 2023-06-14
feat: [PL-39005]: bijou implementation delegate side

xingchi-jin:
## Describe your changes
This change does
1. Acquire task payload using the new API created at PR https://github.com/harness/harness-core/pull/48331 
2. based on different type of request init or exec or clean, it invokes different handlers 
3. Based on runner type, it invokes different runners. 
4. Added a runner factory so we can easily add more runners. defined a common runner interface, which essentially takes binary input

## Checklist
- [ ] I've documented the changes in the PR description.
- [ ] I've tested this change either in PR or local environment.
- [ ] If hashcheck failed I followed [the checklist](https://harness.atlassian.net/wiki/spaces/DEL/pages/21016838831/PR+Codebasehash+Check+merge+checklist) and updated this PR description with my answers to make sure I'm not introducing any breaking changes.

## Comment Triggers
<details>
  <summary>Build triggers</summary>

- Feature build: `trigger feature-build`
- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger events_fmwrk_monitor`
  - `trigger event_server`
  - `trigger change_data_capture`
  -  Trigger multiple builds together: For eg: `trigger dms, manager`
- Immutable delegate `trigger publish-delegate`
</details>

## [Latest PR Check Triggers](https://github.com/harness/harness-core/blob/develop/.github/pull_request_template.md)

<details>
  <summary>PR Check triggers</summary>
You can run multiple PR check triggers by comma separating them in a single comment. e.g. `trigger ut0, ut1`

- Compile: `trigger compile`
- CodeformatCheckstyle: `trigger checkstylecodeformat`
  - CodeFormat: `trigger codeformat`
  - Checkstyle: `trigger checkstyle`
- MessageMetadata: `trigger messagecheck`
- File-Permission-Check: `trigger checkpermission`
- Recency: `trigger recency`
- BuildNumberMetadata: `trigger buildnum`
- Trigger CommonChecks: `trigger commonchecks`
- PMD: `trigger pmd`
- Copyright Check: `trigger copyrightcheck`
- Feature Name Check: `trigger featurenamecheck`
- UnitTests-ALL: `trigger uts`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
[...]

PR #49775, opened 2023-06-26
feat: [PL-39005]: runner factory, skeleton

xingchi-jin:
## Describe your changes
Split from https://github.com/harness/harness-core/pull/49272 
This part contains runner factory, runner interface. It's not plugged into the runtime flow. 

## Checklist

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/49775)


260-delegate/src/main/java/io/harness/delegate/service/runners/RunnersFactory.java
```diff
@@ -0,0 +1,45 @@
...
+
+import static io.harness.delegate.beans.RunnerType.RUNNER_TYPE_K8S;
+
+@Slf4j
+@Singleton
+public final class RunnersFactory {
+    private final Map<String, Runner> runnerRegistry = new HashMap<>();
+
+    @Inject
+    public RunnersFactory(@Named("k8sRunner") final Runner k8SLiteRunner) {
```
gn-harness, line 28: You can directly inject K8sLiteRunner and DockerRunner in future. do we need these Named annotations?

260-delegate/src/main/java/io/harness/delegate/service/core/litek8s/K8SLiteRunner.java
```diff
@@ -59,7 +60,7 @@ public class K8SLiteRunner implements TaskRunner {
   //  private final K8EventHandler k8EventHandler;
 
   @Override
-  public void init(final String taskGroupId, final InputData infra) {
+  public void init(final String taskGroupId, final InputData infra, final Context context) {
```
gn-harness, line 63: Context would be used in the next PR?

260-delegate/src/main/java/io/harness/delegate/service/runners/itfc/Runner.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+package io.harness.delegate.service.runners.itfc;
+
+import io.harness.delegate.core.beans.InputData;
+import io.harness.delegate.service.handlermapping.context.Context;
+
+public interface Runner {
+  void init(String infraId, InputData infra, Context context);
+  void execute(String infraId, InputData tasks, Context context);
```
gn-harness, line 15: Some documentation on these types would help, especially Context

260-delegate/src/main/java/io/harness/delegate/service/core/litek8s/K8SLiteRunner.java
xingchi-jin: yes, the context passes information obtainable only from delegate or manager, not from the APIs (which ci, cd calls), for example, task id, decrypted values.

260-delegate/src/main/java/io/harness/delegate/service/runners/itfc/Runner.java
xingchi-jin: got it

260-delegate/src/main/java/io/harness/delegate/service/runners/RunnersFactory.java
xingchi-jin: i added a named annotation because we could have more types of classes inherit the Runner interface. For example, when we add VM runner, we need some VM runner client that talks to VM runner.
So in this case, adding a named annotation helps differentiate between different runner implementations

mdmilic (changes requested):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49775)* status: 16 file(s) to review. 1 change requests. 3. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/app/modules/platform/k8s/K8SRunnerModule.java` line 31 at r1](https://reviewable.io/reviews/harness/harness-core/49775#-NYx6ThF7eJZ23BP1jx1:-NYx6ThG-ey4_B-GfMhV:b-ugj408) ([raw file](https://github.com/harness/harness-core/blob/6ec13f629764221d434445246691ef9ef411783a/260-delegate/src/main/java/io/harness/delegate/app/modules/platform/k8s/K8SRunnerModule.java#L31)):*
> ```Java
> 
>     bind(K8SRunnerConfig.class).toInstance(config);
>     bind(Runner.class).annotatedWith(Names.named("k8sRunner")).to(K8SLiteRunner.class);
> ```

We should have only single default runner (e.g. VM/Exec runners are not baked into the delegate), we can add named later on if really needed, but I don't see a reason for it right now.
___
*[`260-delegate/src/main/java/io/harness/delegate/service/core/litek8s/K8SLiteRunner.java` line 63 at r1](https://reviewable.io/reviews/harness/harness-core/49775#gh-1242957778:-NYx6yZU0kIPb9C9tvVm:b-dqkk28) ([raw file](https://github.com/harness/harness-core/blob/6ec13f629764221d434445246691ef9ef411783a/260-delegate/src/main/java/io/harness/delegate/service/core/litek8s/K8SLiteRunner.java#L63)):*

yes, the context passes information obtainable only from delegate or manager, not from the APIs (which ci, cd calls), for example, task id, decrypted values. 
</blockquote></details>

Most of the fields you have in context are static, they never change throughout the delegate lifecycle. There is no need to pass them around with every request. Instead all this info should be given to the runner when you are starting it up (e.g. guice injected for K8S Runner).  We already have it in the K8SRunnerConfig (though you could make that more flexible and serializable to generally applies to all runners. 
Other things like taskId doesn't really make sense here as the Context doesn't/shouldn't really know any specifics of the infra. Apart from that you already have `taskGroupId` (which is ID for the entire transaction), but also `taskId` (which is ID for each specific step) inside the infra spec right now (we can pass it around differently, but still needs to be a collection of values and to have mapping between infra and ID)
___
*[`260-delegate/src/main/java/io/harness/delegate/service/runners/RunnersFactory.java` line 28 at r1](https://reviewable.io/reviews/harness/harness-core/49775#gh-1242956881:-NYxMnJV3ySdX_CnMlEo:b-be134d) ([raw file](https://github.com/harness/harness-core/blob/6ec13f629764221d434445246691ef9ef411783a/260-delegate/src/main/java/io/harness/delegate/service/runners/RunnersFactory.java#L28)):*

i added a named annotation because we could have more types of classes inherit the Runner interface. For example, when we add VM runner, we need some VM runner client that talks to VM runner.
So in this case, adding a named annotation helps differentiate between different runner implementations 
</blockquote></details>

Let's drop named for now as I don't think this would be a use case for now (there could be other runners implementing the interface, but not bundled within delegate anyway for now). If it becomes a thing we will easily add it. 
Also, I think there is a fundamental misunderstanding how runners would work, the only runner related thing that will be bundled into the delegate executable is going to be `K8SLiteRunner`, there is not going to be any clients, wrappers or any specific things related to other runners (VM or otherwise). There can be an HTTP/GRPC client which can communicate with various runners using runner protocol we define, but this client will not be runner specific, it should use generic runner protocol and be able to communicate with any runner given the endpoint. Also, if we decide, the communication can be backwards only (i.e. runner -> delegate core) which is what drone uses for some runner types or we can support both (TBD, not part of this PR). You need to think of other runners as plugins as well which are just configurable/discoverable by delegate, so you can drop in any runner which conforms to the standards we define. 
So just to continue the discussion [from here](https://reviewable.io/reviews/harness/harness-core/49272#-NY_2QOe39YSoCLb0Oem), I'm not exactly sure how this would work for other runner types and if needed at all

xingchi-jin (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49775)* status: 16 file(s) to review. 1 change requests. 4. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/service/core/litek8s/K8SLiteRunner.java` line 63 at r1](https://reviewable.io/reviews/harness/harness-core/49775#gh-1242957778:-NYxlLsx-6XgWU9p9ePI:b-p8bfrr) ([raw file](https://github.com/harness/harness-core/blob/6ec13f629764221d434445246691ef9ef411783a/260-delegate/src/main/java/io/harness/delegate/service/core/litek8s/K8SLiteRunner.java#L63)):*

Most of the fields you have in context are static, they never change throughout the delegate lifecycle. There is no need to pass them around with every request. Instead all this info should be given to the runner when you are starting it up (e.g. guice injected for K8S Runner).  We already have it in the K8SRunnerConfig (though you could make that more flexible and serializable to generally applies to all runners. 
Other things like taskId doesn't really make sense here as the Context doesn't/shouldn't really know any specifics of the infra. Apart from that you already have `taskGroupId` (which is ID for the entire transaction), but also `taskId` (which is ID for each specific step) inside the infra spec right now (we can pass it around differently, but still needs to be a collection of values and to have mapping between infra and ID)
</blockquote></details>

i can put the static ones into k8sconfig and pass through injection. 
I need context object to pass additional runtime information to runner interface 
___
*[`260-delegate/src/main/java/io/harness/delegate/service/runners/RunnersFactory.java` line 28 at r1](https://reviewable.io/reviews/harness/harness-core/49775#gh-1242956881:-NYxl83-6GtgKbybmBMX:b-5dea6e) ([raw file](https://github.com/harness/harness-core/blob/6ec13f629764221d434445246691ef9ef411783a/260-delegate/src/main/java/io/harness/delegate/service/runners/RunnersFactory.java#L28)):*

Let's drop named for now as I don't think this would be a use case for now (there could be other runners implementing the interface, but not bundled within delegate anyway for now). If it becomes a thing we will easily add it. 
Also, I think there is a fundamental misunderstanding how runners would work, the only runner related thing that will be bundled into the delegate executable is going to be `K8SLiteRunner`, there is not going to be any clients, wrappers or any specific things related to other runners (VM or otherwise). There can be an HTTP/GRPC client which can communicate with various runners using runner protocol we define, but this client will not be runner specific, it should use generic runner protocol and be able to communicate with any runner given the endpoint. Also, if we decide, the communication can be backwards only (i.e. runner -> delegate core) which is what drone uses for some runner types or we can support both (TBD, not part of this PR). You need to think of other runners as plugins as well which are just configurable/discoverable by delegate, so you can drop in any runner which conforms to the standards we define. 
So just to continue the discussion [from here](https://reviewable.io/reviews/harness/harness-core/49272#-NY_2QOe39YSoCLb0Oem), I'm not exactly sure how this would work for other runner types and if needed at all
</blockquote></details>

i can remove it now. but it's minor internal implementation. i don't think there is any misunderstanding, it's more about different ways if implementing code

gn-harness (dismissed):

mdmilic (dismissed):
> Reviewed 1 of 2 files at r2, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49775)* status: 20 file(s) to review. 3. unresolved discussions. 1 of 2 approvals obtained. 

___
*[`260-delegate/container/scripts/delegate.sh` line 153 at r3](https://reviewable.io/reviews/harness/harness-core/49775#-NYyw4y34w9ccFdxdbgD:-NYyw4y49t4MydelCBtD:bp4vxir) ([raw file](https://github.com/harness/harness-core/blob/02be7cced1339b5291c8b6280ccf9186cf82f530/260-delegate/container/scripts/delegate.sh#L153)):*
> ```Shell
> if [ ! -e config-delegate.yml ]; then
>   echo "accountId: $ACCOUNT_ID" > config-delegate.yml
>   echo "delegateName: $DELEGATE_NAME" > config-delegate.yml
> ```

FYI this is legacy delegate script, not used for immutable.

xingchi-jin (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49775)* status: 20 file(s) to review. 3. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`260-delegate/src/main/java/io/harness/delegate/service/runners/RunnersFactory.java` line 28 at r1](https://reviewable.io/reviews/harness/harness-core/49775#gh-1242956881:-NYy3oZk9vdWUgOw7Sv-:b-w1t1vn) ([raw file](https://github.com/harness/harness-core/blob/6ec13f629764221d434445246691ef9ef411783a/260-delegate/src/main/java/io/harness/delegate/service/runners/RunnersFactory.java#L28)):*

i can remove it now. but it's minor internal implementation. i don't think there is any misunderstanding, it's more about different ways if implementing code 
</blockquote></details>

resolve this, we discussed in meeting, we are having same understanding 
___
*[`260-delegate/container/scripts/delegate.sh` line 153 at r3](https://reviewable.io/reviews/harness/harness-core/49775#-NYyw4y34w9ccFdxdbgD:-NYz7p_TCHId8xJDrDhX:b-5dfryo) ([raw file](https://github.com/harness/harness-core/blob/02be7cced1339b5291c8b6280ccf9186cf82f530/260-delegate/container/scripts/delegate.sh#L153)):*

FYI this is legacy delegate script, not used for immutable.
</blockquote></details>

removed it

xingchi-jin:
trigger uts

xingchi-jin:
trigger uts

xingchi-jin:
trigger uts

xingchi-jin:
trigger uts

xingchi-jin:
trigger ut9

Raghav-Murali (approved):

mdmilic (approved):
> Reviewed 5 of 5 files at r4, 1 of 2 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/49775)* status: 20 file(s) to review. 3. unresolved discussions.

PR #49985, opened 2023-06-30
feat: [CI-1382]: add api handler to get list based on prefix limited to 100

rafaelmart1ns:
## Checklist

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/49985)


product/log-service/cache/cache.go
```diff
@@ -0,0 +1,21 @@
...
+	Create(context.Context, string, interface{}) error
+
+	// Get an entry into log
+	Get(context.Context, string) ([]byte, error)
+
+	// Ping the cache for readiness
+	Ping(context.Context) error
+
+	// Exists file in bucket
+	Exists(context.Context, string) bool
```
vistaarjuneja, line 17: Remove exists - instead return `exists` as part of `Get` return

product/log-service/cache/cache.go
```diff
@@ -0,0 +1,21 @@
...
+
+import "context"
+
+// Cache defines the cache functions interface
+type Cache interface {
+	// Create an entry into log
+	Create(context.Context, string, interface{}) error
+
+	// Get an entry into log
+	Get(context.Context, string) ([]byte, error)
```
vistaarjuneja, line 11: `Get` can return `interface{}, bool` where bool will be exists or not

product/log-service/cache/memory/memory.go
```diff
@@ -0,0 +1,56 @@
...
+	"sync"
+
+	"github.com/harness/harness-core/product/log-service/cache"
+)
+
+var _ cache.Cache = (*Cacher)(nil)
+
+type Cacher struct {
+	sync.Mutex
+	caches map[string]interface{}
```
vistaarjuneja, line 15: Nit: caches -> cache

product/log-service/cache/memory/memory.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func New() *Cacher {
+	return &Cacher{
+		caches: make(map[string]interface{}),
+	}
+}
+
+func (c Cacher) Create(ctx context.Context, key string, i interface{}) error {
+	c.Lock()
```
vistaarjuneja, line 25: Let's call `defer c.Unlock()` right after `c.Lock()`.

product/log-service/db/redis/redis.go
```diff
@@ -0,0 +1,149 @@
+package redis
```
vistaarjuneja, line 1: renaming files is better than moving the code since that helps to see only the diff which has been made. can you try to see if we can do that instead? will help with the review

product/log-service/handler/middleware.go
```diff
@@ -119,3 +127,112 @@ func doApiKeyAuthentication(inputApiKey, accountID, routingId string, ngClient *
...
+				return
+			}
+		})
+	}
+}
+
+func ValidatePrefixRequest() func(handler http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			containRunSequence := strings.Contains(r.URL.String(), "runSequence")
```
vistaarjuneja, line 217: we should not have any logic in log service which tries to look at the log key

product/log-service/store/bolt/bolt.go
```diff
@@ -114,3 +114,11 @@ func (s *Store) Delete(ctx context.Context, key string) error {
 func (s *Store) Ping() error {
 	return nil
 }
+
+func (s *Store) ListBlobPrefix(ctx context.Context, prefix string) (map[int][]string, error) {
+	return nil, errors.New("not implemented, use the s3 Datastore")
+}
+
+func (s *Store) Finish(ctx context.Context, key string) (bool, error) {
```
vistaarjuneja, line 122: can you add comments as to what Finish is

product/log-service/store/bolt/bolt.go
```diff
@@ -114,3 +114,11 @@ func (s *Store) Delete(ctx context.Context, key string) error {
 func (s *Store) Ping() error {
 	return nil
 }
+
+func (s *Store) ListBlobPrefix(ctx context.Context, prefix string) (map[int][]string, error) {
```
vistaarjuneja, line 118: add one line comments explaining what the function does

product/log-service/store/s3/s3.go
```diff
@@ -9,69 +9,142 @@ package s3
...
+	}
+}
+
+type RecalculateV4Signature struct {
+	next   http.RoundTripper
+	signer *v4.Signer
+	cfg    aws.Config
+}
+
+func (lt *RecalculateV4Signature) RoundTrip(req *http.Request) (*http.Response, error) {
```
vistaarjuneja, line 108: What does this function do?

vistaarjuneja (commented):
first round of comments, will review more

product/log-service/cache/cache.go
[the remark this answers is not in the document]
rafaelmart1ns: Yes, because redis return a string, and transforming it to a byte is less costly than to interface

product/log-service/cache/memory/memory.go
rafaelmart1ns: done.

product/log-service/cache/memory/memory.go
rafaelmart1ns: It was changed.

product/log-service/handler/middleware.go
rafaelmart1ns: we need this small validation to prevent the misuse of the prefix, because if the client passes a prefix without the runsequence it indicates that it will get log data for the entire repository which prevents the client from intercepting the request and sending it and downloading it all his logs and not just one run.

product/log-service/store/bolt/bolt.go
rafaelmart1ns: ditto.

product/log-service/cache/cache.go
rafaelmart1ns: Redis supports exists to use less memory, so I don't store the values returned by get to check if it exists, this will save memory.

product/log-service/handler/entity.go
```diff
@@ -0,0 +1,36 @@
...
+// https://polyformproject.org/wp-content/uploads/2020/05/PolyForm-Free-Trial-1.0.0.txt.
+
+package handler
+
+import (
+	"encoding/json"
+	"time"
+)
+
+type Info struct {
```
vistaarjuneja, line 13: let's reuse the `Link` struct declared in log service/client/client.go and add any information as part of another struct. Also, rename `Info` to be something more clear

product/log-service/handler/handler.go
```diff
@@ -124,7 +127,25 @@ func Handler(stream stream.Stream, store store.Store, config config.Config, ngCl
...
 		sr := chi.NewRouter()
-		sr.Get("/", HandlePing(stream, store))
+		sr.Get("/", HandlePing(cache, stream, store))
+
+		return sr
+	}())
+
+	// Blob zip store endpoints
+	// Format: /blob/prefix?accountID=&prefix=&pipeline_name=&execution_id=
+	r.Mount("/blob/prefix", func() http.Handler {
```
vistaarjuneja, line 137: Can we change this API a bit and make it `/blob/zip?accountID=<>&prefix=<>`. Let's communicate this to UI as well

product/log-service/handler/middleware.go
```diff
@@ -119,3 +129,119 @@ func doApiKeyAuthentication(inputApiKey, accountID, routingId string, ngClient *
...
+
+func CacheRequest(cache cache.Cache, s store.Store) func(handler http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			logger.WithContext(context.Background(), logger.FromRequest(r))
+
+			var info Info
+			prefix := r.URL.Query().Get("prefix")
+			inf, err := cache.Get(context.Background(), prefix)
+			if err != nil && err != redis.Nil {
```
vistaarjuneja, line 141: this should be || not &&

product/log-service/handler/middleware.go
```diff
@@ -119,3 +129,119 @@ func doApiKeyAuthentication(inputApiKey, accountID, routingId string, ngClient *
...
+				logger.FromRequest(r).
+					WithError(err).
+					WithField("url", r.URL.String()).
+					WithField("time", time.Now().Format(time.RFC3339)).
+					Errorln("middleware cache: failed to get cache")
+			}
+
+			if len(inf) > 0 {
```
vistaarjuneja, line 149: why are we continuing here if err != nil? Do we need to check all this? can't we just return after L146

product/log-service/handler/middleware.go
```diff
@@ -119,3 +129,119 @@ func doApiKeyAuthentication(inputApiKey, accountID, routingId string, ngClient *
...
 	return err
 }
+
+func CacheRequest(cache cache.Cache, s store.Store) func(handler http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			logger.WithContext(context.Background(), logger.FromRequest(r))
+
+			var info Info
+			prefix := r.URL.Query().Get("prefix")
```
vistaarjuneja, line 139: Don't hardcode "prefix". We have `prefixParam` declared, you can reuse that

product/log-service/handler/store.go
```diff
@@ -172,3 +181,213 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+				logger.FromRequest(r).
+					WithError(err).
+					WithField("url", r.URL.String()).
+					WithField("time", time.Now().Format(time.RFC3339)).
+					Errorln("ziplog_start: failed to create cache")
+			}
+			go zipUploadRoutine(s, c, *r, zipPrefix, prefix, out, info, wp)
+		}
+
+		WriteJSON(w, info, 200)
```
vistaarjuneja, line 242: If exists is true, this will just write an empty Info in the response - is that expected?

product/log-service/handler/store.go
```diff
@@ -172,3 +181,213 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+		if err != nil || len(out) == 0 {
+			WriteNotFound(w, err)
+			logger.FromRequest(r).
+				WithError(err).
+				WithField("prefix", prefix).
+				Errorln("api: cannot list files blob to download")
+			return
+		}
+
+		uid := uuid.New().String()
```
vistaarjuneja, line 208: why are we generating a new UID each time? Let's just use <prefix>_logs.zip to store the logs?

product/log-service/store/s3/s3.go
```diff
@@ -147,3 +148,44 @@ func (s *Store) Ping() error {
...
+	})
+
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+func (s *Store) ListBlobPrefix(ctx context.Context, prefix string) (map[int][]string, error) {
```
vistaarjuneja, line 166: why is this a map[int][]string? Can't we just return a []string with the keys

product/log-service/handler/store.go
```diff
@@ -172,3 +181,213 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+			WithField("latency", time.Since(st)).
+			WithField("time", time.Now().Format(time.RFC3339)).
+			Infoln("api: successfully list prefix finished")
+	}
+
+}
+
+func zipUploadRoutine(s store.Store, c cache.Cache, r http.Request, zipPrefix, prefix string, out map[int][]string, info Info, wp *worker.WorkerPool) {
+	var wg sync.WaitGroup
+	internal, cancel := context.WithCancel(context.Background())
```
vistaarjuneja, line 254: rename internal -> ctx

product/log-service/handler/store.go
```diff
@@ -172,3 +181,213 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+			}
+
+			logger.FromRequest(&r).
+				WithField("prefix", prefix).
+				WithField("time", time.Now().Format(time.RFC3339)).
+				Debug("ziplog: successfully zip created")
+		})
+	}
+
+	go func() {
```
vistaarjuneja, line 352: this need not be another goroutine - we are waiting for the upload to complete before returning right?

product/log-service/handler/store.go
```diff
@@ -172,3 +181,213 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+		WithField("prefix", prefix).
+		WithField("time", time.Now().Format(time.RFC3339)).
+		Debug("ziplog: starting zip and upload log files")
+
+	pipeRead, pipeWrite := io.Pipe()
+
+	zipWriter := zip.NewWriter(pipeWrite)
+
+	// upload
+	go func() {
```
vistaarjuneja, line 267: this goroutine should be part of a wait group as well since we want to wait for upload to complete before returning

product/log-service/handler/store.go
```diff
@@ -172,3 +181,213 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+			WithField("prefix", prefix).
+			WithField("latency", time.Since(st)).
+			WithField("time", time.Now().Format(time.RFC3339)).
+			Infoln("api: successfully list prefix finished")
+	}
+
+}
+
+func zipUploadRoutine(s store.Store, c cache.Cache, r http.Request, zipPrefix, prefix string, out map[int][]string, info Info, wp *worker.WorkerPool) {
+	var wg sync.WaitGroup
```
vistaarjuneja, line 253: use errorgroup.Group. We need error propagation and cancellation here

product/log-service/handler/store.go
```diff
@@ -172,3 +181,213 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+	internal, cancel := context.WithCancel(context.Background())
+	logger.WithContext(internal, logger.FromRequest(&r))
+
```
vistaarjuneja, line 256:
```
g := new(errgroup.Group)
g.Go(routine to write into the writer)
g.Go(routine to upload to s3)
err := g.Wait()
if err != nil {
    // handle error
}
write to cache
```

vistaarjuneja: you can use something similar to this ^

product/log-service/handler/store.go
```diff
@@ -172,3 +181,213 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+			WithField("prefix", prefix).
+			WithField("time", time.Now().Format(time.RFC3339)).
+			Debug("ziplog: successfully zip file updated to s3")
+	}()
+
+	var syncZip sync.Mutex
+	// zip
+	for _, keys := range out {
+		keys := keys
+		wg.Add(1)
```
vistaarjuneja, line 314: create the worker pool here itself instead of using a global pool. We can initialise the pool with x parallel executors and keep the length of the channel as x as well. (x can be equal to 5)

product/log-service/store/s3/s3.go
```diff
@@ -147,3 +148,44 @@ func (s *Store) Ping() error {
...
+	keys := make(map[int][]string)
+
+	params := &s3.ListObjectsInput{
+		Bucket:  aws.String(s.bucket),
+		Prefix:  aws.String(prefix),
+		MaxKeys: aws.Int64(100),
+	}
+
+	pageNum := 0
+	err := svc.ListObjectsPages(params, func(p *s3.ListObjectsOutput, lastPage bool) bool {
```
vistaarjuneja, line 178: will this list  100 pages for each pageNum? We should keep 100 pages in total (that should be the max limit on the total number of keys)

product/log-service/handler/store.go
```diff
@@ -172,3 +181,213 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+			}
+			err := c.Create(ctx, prefix, info)
+			if err != nil {
+				logger.FromRequest(r).
+					WithError(err).
+					WithField("url", r.URL.String()).
+					WithField("time", time.Now().Format(time.RFC3339)).
+					Errorln("ziplog_start: failed to create cache")
+			}
+			go zipUploadRoutine(s, c, *r, zipPrefix, prefix, out, info, wp)
```
vistaarjuneja, line 239: we should use the global worker pool at this level - can be done in next PR. We can sync up about this

rafaelmart1ns:
trigger manager

rafaelmart1ns:
trigger ng_manager

product/log-service/handler/store.go
```diff
@@ -172,3 +179,74 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+				WithField("info", info).
+				Errorln("api: cannot create cache info")
+			WriteInternalError(w, err)
+			return
+		}
+
+		// produce event to queue
+		err = q.Produce(ctx, cfg.ConsumerWorker.StreamName, prefix, zipPrefix, out)
+		if err != nil {
+			WriteNotFound(w, err)
```
lasdox, line 236: Why NotFound? Should be InternalError too I think

product/log-service/queue/redis/redis.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (r Redis) Produce(ctx context.Context, stream, key, zipKey string, values map[int][]string) error {
+	_, err := r.Client.XAdd(ctx, &redis.XAddArgs{
+		Stream: stream,
+		Values: map[string]interface{}{"prefix": entity.EventQueue{key, zipKey, values}},
+	}).Result()
+	if err != nil {
+		fmt.Println("error to add message to redis")
+		return err
```
lasdox, line 40: need to remove all these fmt.Print

lasdox: lets not log anything here, caller should be logging.
We can add a wrapped error to give more context

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,303 @@
...
+				logEntryWorker.
+					WithField("time", time.Now().Format(time.RFC3339)).
+					Infoln("consumer-initiated listening")
+
+				message, err := wp.queue.Consume(ctx, wp.streamName, wp.consumerGroup, wID)
+				if err != nil {
+					logEntryWorker.
+						WithField("time", time.Now().Format(time.RFC3339)).
+						Errorln("consumer execute: cannot process message")
+					continue
```
lasdox, line 88: should we add some sleep before continuing?

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,303 @@
...
+	var syncZip sync.Mutex
+
+	pipeRead, pipeWrite := io.Pipe()
+	zipWriter := zip.NewWriter(pipeWrite)
+
+	// upload
+	go func() {
+		select {
+		case <-ctx.Done():
+			return
```
lasdox, line 237: you dont need to wait for ctx cancelation by upload side, this will be canceled by aws sdk

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,303 @@
...
+				return
+			}
+			break
+		default:
+		}
+	}
+}
+
+func zipUploadRoutine(ctx context.Context, s store.Store, zipPrefix string, out map[int][]string, errs chan error, finishes chan string) {
+	var wg sync.WaitGroup
```
lasdox, line 227: I dont think this function should be coupled to consumer, this is business logic
instead consumer should be able to use the zipUpload
this seems like an inverted dependency

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,303 @@
...
+	go func() {
+		defer wg.Done()
+
+		for _, keys := range out {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+			zipFile(ctx, keys, zipWriter, s, &syncZip, errs)
```
lasdox, line 259: nit: you can put zipfile on the default block

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+func zipFile(ctx context.Context, keys []string, zipWriter *zip.Writer, s store.Store, syncZip *sync.Mutex, errs chan error) {
+	for _, key := range keys {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
+		fileDownloaded, err := s.Download(ctx, key)
```
lasdox, line 279: I dont think download should happen inside a zipFile function

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,303 @@
...
+				}
+
+				err = json.Unmarshal(infoBytes, &info)
+				if err != nil {
+					logEntry.
+						WithError(err).
+						WithField("step", s).
+						WithField("time", time.Now().Format(time.RFC3339)).
+						Errorln("polling zip routine: cannot get cache")
+					return
```
lasdox, line 205: polling zip routine: cannot unmarshal

product/log-service/db/redis/redis.go
rafaelmart1ns: as discussed sync it not possible. closing it.

product/log-service/handler/handler.go
rafaelmart1ns: it was changed.

product/log-service/handler/middleware.go
rafaelmart1ns: it was changed.

product/log-service/handler/store.go
rafaelmart1ns: yes, it was changed.

product/log-service/queue/redis/redis.go
rafaelmart1ns: it was changed.

product/log-service/store/s3/s3.go
```diff
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/credentials"
+
```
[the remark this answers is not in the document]
rafaelmart1ns: I used it and it return this space.

product/log-service/store/s3/s3.go
```diff
@@ -147,3 +148,44 @@ func (s *Store) Ping() error {
 
 	return nil
 }
+
+func (s *Store) Finish(ctx context.Context, key string) (bool, error) {
+	svc := s3.New(s.session)
+	_, err := svc.HeadObject(&s3.HeadObjectInput{
```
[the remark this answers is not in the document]
rafaelmart1ns: In was removed.

product/log-service/store/s3/s3.go
rafaelmart1ns: You are right, it was changed.

product/log-service/store/s3/s3.go
rafaelmart1ns: it was changed to use env var to set it.

product/log-service/worker/consumer/consumer.go
rafaelmart1ns: it was changed

product/log-service/worker/consumer/consumer.go
rafaelmart1ns: it was changed

product/log-service/worker/consumer/consumer.go
rafaelmart1ns: it is not necessary because the consume sleep the execution waiting for a message both in the redis and in-memory implementations.

product/log-service/worker/consumer/consumer.go
rafaelmart1ns: it was changed

product/log-service/handler/store.go
rafaelmart1ns: It was changed adding the queue using redis and in memory

product/log-service/handler/handler.go
```diff
@@ -124,7 +126,23 @@ func Handler(stream stream.Stream, store store.Store, config config.Config, ngCl
...
 	r.Mount("/ready/healthz", func() http.Handler {
 		sr := chi.NewRouter()
-		sr.Get("/", HandlePing(stream, store))
+		sr.Get("/", HandlePing(cache, stream, store))
+
+		return sr
+	}())
+
+	// Blob zip store endpoints
+	// Format: /blob/prefix?accountID=&prefix=&pipeline_name=&execution_id=
```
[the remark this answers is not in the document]
rafaelmart1ns: It was changed.

product/log-service/handler/entity.go
rafaelmart1ns: It was changed

product/log-service/handler/middleware.go
rafaelmart1ns: it doesn't make sense anymore.

product/log-service/handler/middleware.go
rafaelmart1ns: it doesn't make sense anymore.

product/log-service/handler/store.go
rafaelmart1ns: no, it was fixed.

product/log-service/cache/redis/redis.go
```diff
@@ -0,0 +1,77 @@
...
+	stringCmd, err := r.Client.Get(key).Result()
+	if err != nil {
+		return []byte{}, err
+	}
+	return []byte(stringCmd), nil
+}
+
+func (r *Redis) Create(ctx context.Context, key string, value interface{}) error {
+	r.Client.Del(key)
+	cmd := r.Client.Set(key, value, time.Hour)
```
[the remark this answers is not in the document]
rafaelmart1ns: it was changed to use env var.

product/log-service/db/redis/redis.go
```diff
@@ -0,0 +1,155 @@
...
+	roots := x509.NewCertPool()
+	ok := roots.AppendCertsFromPEM(rootPem)
+	if !ok {
+		return nil, fmt.Errorf("error adding cert (%s) to pool, error: %s", certPathForTLS, err.Error())
+	}
+	return &tls.Config{RootCAs: roots}, nil
+}
+
+// Helper function to set an expiry to a key if it's not already set
+func setExpiry(ctx context.Context, key string, expiry time.Duration, client *redis.Client) error {
```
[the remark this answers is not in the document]
rafaelmart1ns: Yes it was changed.

product/log-service/handler/middleware.go
```diff
@@ -119,3 +129,119 @@ func doApiKeyAuthentication(inputApiKey, accountID, routingId string, ngClient *
...
+			inf, err := cache.Get(context.Background(), prefix)
+			if err != nil && err != redis.Nil {
+				logger.FromRequest(r).
```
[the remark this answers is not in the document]
rafaelmart1ns: it was changed.

product/log-service/store/s3/s3.go
```diff
@@ -147,3 +148,43 @@ func (s *Store) Ping() error {
...
+}
+
+func (s *Store) ListBlobPrefix(ctx context.Context, prefix string) (map[int][]string, error) {
+	svc := s3.New(s.session)
+
+	keys := make(map[int][]string)
+
+	params := &s3.ListObjectsInput{
+		Bucket: aws.String(s.bucket),
+		Prefix: aws.String(prefix),
```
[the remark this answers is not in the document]
rafaelmart1ns: it was changed to use envvar.

product/log-service/handler/store.go
rafaelmart1ns: I added a queue on redis with worker pool configured by env var.

product/log-service/handler/store.go
rafaelmart1ns: I updated the code.

product/log-service/handler/store.go
rafaelmart1ns: it was changed

product/log-service/handler/store.go
```diff
@@ -172,3 +179,74 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+		h := w.Header()
+		h.Set("Access-Control-Allow-Origin", "*")
+		ctx := r.Context()
+
+		accountID := r.URL.Query().Get(accountIDParam)
+		prefix := r.URL.Query().Get(UsePrefixParam)
+
+		zipPrefix := uuid.New().String() + "_logs.zip"
+
+		link, err := s.DownloadLink(ctx, zipPrefix, cfg.CacheTTL)
```
smjt-h, line 195: What if the object for the key zipPrefix does not exist in the s3? It will not be able to create a signed download link

rafaelmart1ns: This method creates the promise of the zip file and the signed link, even if the object does not exist the link is signed, and once the upload happens to that name the link will work normally.

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,229 @@
...
+				return fmt.Errorf("zipfile: failed to close: %w", err)
+			}
+			syncZip.Unlock()
+		}
+		return nil
+	})
+
+	err := gErrGroupDownload.Wait()
+	if err != nil {
+		return err
```
vistaarjuneja, line 218: do we need to handle some kind of cleanup here? what happens if one of these ops fail - should we clear the cache and delete the blob entry so that we have a clean slate for the next time this op comes?
We should not leave things in partial state.

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,229 @@
...
+		cache:         c,
+		consumerGroup: consumerGroup,
+		streamName:    streamName,
+		config:        cfg,
+	}
+
+	return wp
+}
+
+func (wp *ConsumerPool) Execute() {
```
vistaarjuneja, line 64: This function is not generic - so let it either accept a function as an arg or move it to a separate folder

product/log-service/handler/helper.go
```diff
@@ -14,7 +14,7 @@ const (
 	accountIDParam = "accountID"
 	keyParam       = "key"
 	snapshotParam  = "snapshot"
-	usePrefixParam = "prefix"
+	UsePrefixParam = "prefix"
```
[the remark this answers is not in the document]
rafaelmart1ns: it was changed

product/log-service/handler/middleware.go
```diff
@@ -119,3 +126,117 @@ func doApiKeyAuthentication(inputApiKey, accountID, routingId string, ngClient *
...
+			exists := c.Exists(ctx, prefix)
+
+			if exists {
+				inf, err := c.Get(ctx, prefix)
+				if err != nil {
+					logger.FromRequest(r).
+						WithError(err).
+						WithField("url", r.URL.String()).
+						WithField("time", time.Now().Format(time.RFC3339)).
+						Errorln("middleware cache: cannot get c")
```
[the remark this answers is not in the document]
rafaelmart1ns: it was changed.

product/log-service/handler/middleware.go
```diff
@@ -119,3 +126,117 @@ func doApiKeyAuthentication(inputApiKey, accountID, routingId string, ngClient *
...
+					return
+				}
+
+				switch info.Status {
+				case entity.QUEUED:
+				case entity.IN_PROGRESS:
+					WriteJSON(w, info, 200)
+					return
+				case entity.ERROR:
+					err := c.Delete(ctx, prefix)
```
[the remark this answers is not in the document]
rafaelmart1ns: The flow returns the error, on the next attempt it will try to zip again. I don't think it's necessary to remove from s3, since it's streaming the vi pipe from the files we need to check at that point and not here.

product/log-service/handler/store.go
```diff
@@ -172,3 +179,74 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		st := time.Now()
+		h := w.Header()
+		h.Set("Access-Control-Allow-Origin", "*")
+		ctx := r.Context()
+
+		accountID := r.URL.Query().Get(accountIDParam)
+		prefix := r.URL.Query().Get(UsePrefixParam)
+
+		zipPrefix := uuid.New().String() + "_logs.zip"
```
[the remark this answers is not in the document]
rafaelmart1ns: No, It was changed to use the same name generate by sha.

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+}
+
+func downloadZipUploadRoutine(ctx context.Context, s store.Store, zipPrefix string, out []string) error {
+	var syncZip sync.Mutex
+
+	pipeRead, pipeWrite := io.Pipe()
+	zipWriter := zip.NewWriter(pipeWrite)
+
+	gErrGroupDownload := new(errgroup.Group)
```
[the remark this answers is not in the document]
rafaelmart1ns: Because the steps occur in parallel but one needs to finish before the other. In case the download needs to finish and then the upload will finish and for that I create both.

product/log-service/entity/response_prefix.go
```diff
@@ -0,0 +1,30 @@
...
+// https://polyformproject.org/wp-content/uploads/2020/05/PolyForm-Free-Trial-1.0.0.txt.
+
+package entity
+
+import (
+	"encoding/json"
+	"time"
+)
+
+type ResponsePrefixDownload struct {
```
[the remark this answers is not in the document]
rafaelmart1ns: as aligned synchronously I will not change to use the link.

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// zip
+	gErrGroupDownload.Go(func() error {
+		for _, key := range out {
+			fileDownloaded, err := s.Download(ctx, key)
+			if err != nil {
+				return fmt.Errorf("zipfile: failed to download: %w", err)
+			}
+
+			syncZip.Lock()
```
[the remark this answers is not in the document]
rafaelmart1ns: I checked and do not need it anymore. It was removed.

product/log-service/worker/consumer/consumer.go
rafaelmart1ns: as aligned synchronously I will not change that.

product/log-service/config/config.go
```diff
@@ -6,13 +6,16 @@
...
 	"github.com/kelseyhightower/envconfig"
 )
 
 // Config provides the system configuration.
 type Config struct {
-	Debug bool `envconfig:"LOG_SERVICE_DEBUG_MODE"`
-	Trace bool `envconfig:"LOG_SERVICE_TRACE"`
+	Debug    bool          `envconfig:"LOG_SERVICE_DEBUG_MODE"`
+	Trace    bool          `envconfig:"LOG_SERVICE_TRACE"`
+	CacheTTL time.Duration `envconfig:"LOG_SERVICE_CACHE_TTL" default:"2h"`
```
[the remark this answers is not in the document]
vistaarjuneja: @rafaelmart1ns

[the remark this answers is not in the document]
vistaarjuneja: are we removing the cache every 2 hours? Why is that needed?

product/log-service/queue/memory/memory.go
```diff
@@ -0,0 +1,48 @@
...
+	return &Queuer{
+		queue: make(chan map[string]interface{}, 10),
+	}
+}
+
+func (q Queuer) Create(ctx context.Context, s string, s2 string) error {
+	return nil
+}
+
+func (q Queuer) Produce(ctx context.Context, s string, s2 string, s3 string, m []string) error {
```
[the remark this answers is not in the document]
vistaarjuneja: @rafaelmart1ns

product/log-service/queue/queue.go
```diff
@@ -0,0 +1,16 @@
...
+package queue
+
+import (
+	"context"
+)
+
+type Queue interface {
+	Create(context.Context, string, string) error
+	Produce(context.Context, string, string, string, []string) error
+	Consume(context.Context, string, string, string) (map[string]interface{}, error)
```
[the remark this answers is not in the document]
vistaarjuneja: @rafaelmart1ns

product/log-service/worker/consumer/consumer.go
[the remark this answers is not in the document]
vistaarjuneja: @rafaelmart1ns that is not how it works. Both goroutines will start as soon as you do `.Go()`. So they will be happening together.

product/log-service/worker/zipwork/zipwork.go
```diff
@@ -0,0 +1,174 @@
+package zipwork
```
vistaarjuneja, line 1: don't name this folder `log-service/worker` since it's not generic. Move this to `log-service/download` or something similar

product/log-service/queue/redis/redis.go
```diff
@@ -0,0 +1,74 @@
...
+		Block:    0,
+		NoAck:    true,
+	}).Result()
+
+	if err != nil {
+		return nil, err
+	}
+
+	// after consume delete message from the stream
+	r.Client.XDel(context.Background(), stream, streams[0].Messages[0].ID)
```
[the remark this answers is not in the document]
vistaarjuneja: @rafaelmart1ns just double confirm that entries are not going into the pending entries list. They should not since I see you have `Noack` set as true - but just double check once locally.

product/log-service/stream/redis/redis.go
```diff
@@ -356,7 +309,7 @@ func (r *Redis) Info(ctx context.Context) *stream.Info {
 
 // Helper function to set an expiry to a key if it's not already set
 func (r *Redis) setExpiry(key string, expiry time.Duration) error {
```
vistaarjuneja, line 311: this function will still expire all the other keys you create for cache and redis queue. How are we preventing that?

product/log-service/worker/consumer/consumer.go
```diff
@@ -0,0 +1,55 @@
...
+	wp := &ConsumerPool{
+		workerLimit:   maxWorker,
+		ConsumerGroup: consumerGroup,
+		StreamName:    streamName,
+	}
+
+	return wp
+}
+
+func (wp *ConsumerPool) Execute(fn func(ctx context.Context, wID string, q queue.Queue, c cache.Cache, s store.Store, cfg config.Config), q queue.Queue, c cache.Cache, s store.Store, cfg config.Config) {
```
vistaarjuneja, line 41: This is not generic, just move it to a folder `log-service/download` so we know this is download related code

product/log-service/worker/zipwork/zipwork.go
rafaelmart1ns: it was changed.

product/log-service/worker/consumer/consumer.go
rafaelmart1ns: it was changed

product/log-service/stream/redis/redis.go
rafaelmart1ns: As all queue and cache keys have their tttl defined, the expiry watcher will not have an impact on them.

rafaelmart1ns:
> https://github.com/harness/harness-core/pull/49985#discussion_r1263423221
This is correct according to the documentation and tests done.

rafaelmart1ns:
> are we removing the cache every 2 hours? Why is that needed?
This value is just a default if the env var does not exist. We need to define a value to put in environments.

product/log-service/handler/store.go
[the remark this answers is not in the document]
rafaelmart1ns: It was changed to use logs into prefix folder.

product/log-service/queue/memory/memory.go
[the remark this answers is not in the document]
rafaelmart1ns: It was changed in intreface.

product/log-service/queue/queue.go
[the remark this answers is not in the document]
rafaelmart1ns: It was changed.

product/log-service/config/config.go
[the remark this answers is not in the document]
rafaelmart1ns: https://github.com/harness/harness-core/pull/49985#issuecomment-1636128736
This value is just a default if the env var does not exist. We need to define a value to put in environments.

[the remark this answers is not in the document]
rafaelmart1ns: It was changed to 720h that corresponsd to 30 days.

product/log-service/worker/consumer/consumer.go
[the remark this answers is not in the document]
rafaelmart1ns: Yes, maybe I expressed myself wrong. They start together with a sequential download and upload flow, but the pipe can only be closed when the download is finished and when the pipe is closed and the upload understands that it is finished and ends when the last byte of the pipe is read.

product/log-service/handler/handler.go
```diff
@@ -124,7 +127,24 @@ func Handler(stream stream.Stream, store store.Store, config config.Config, ngCl
...
+		sr := chi.NewRouter()
+		sr.Use(RequiredQueryParams(accountIDParam, usePrefixParam))
+		sr.Use(ValidatePrefixRequest())
+		sr.Use(CacheRequest(cache))
+
+		if !config.Auth.DisableAuth {
+			sr.Use(AuthMiddleware(config, ngClient, true))
+		}
+
+		sr.Get("/", HandleZipLinkPrefix(queue, store, cache, config))
```
vistaarjuneja, line 147: @rafaelmart1ns we should make this a POST call as it's not read-only. And we should get this changed in UI as well. cc: @Sagar-Rajput7

product/log-service/handler/store.go
```diff
@@ -172,3 +177,73 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		st := time.Now()
+		h := w.Header()
+		h.Set("Access-Control-Allow-Origin", "*")
+		ctx := r.Context()
+
+		accountID := r.URL.Query().Get(accountIDParam)
+		prefix := r.URL.Query().Get(usePrefixParam)
+
+		zipPrefix := prefix + "/" + "logs.zip"
```
vistaarjuneja, line 191: move "logs.zip" as a variable - something like `downloadSuffix`

product/log-service/config/config.go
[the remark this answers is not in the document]
rafaelmart1ns: It was changed to one week

product/log-service/handler/handler.go
rafaelmart1ns: As discussed an slack it will changed after the freeze.

product/log-service/queue/redis/redis.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2021 Harness Inc. All rights reserved.
+// Use of this source code is governed by the PolyForm Free Trial 1.0.0 license
+// that can be found in the licenses directory at the root of this repository, also available at
+// https://polyformproject.org/wp-content/uploads/2020/05/PolyForm-Free-Trial-1.0.0.txt.
+
+package redis
```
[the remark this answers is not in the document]
rafaelmart1ns: Redis is more reliable. And the memory implementation works fine too.

product/log-service/handler/store.go
```diff
@@ -172,3 +177,74 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+		st := time.Now()
+		h := w.Header()
+		h.Set("Access-Control-Allow-Origin", "*")
+		ctx := r.Context()
+
+		accountID := r.URL.Query().Get(accountIDParam)
+		prefix := r.URL.Query().Get(usePrefixParam)
+
+		filepathSuffix := "logs.zip"
+		zipPrefix := accountID + "/" + prefix + "/" + filepathSuffix
```
vistaarjuneja, line 192: Use this function `CreateAccountSeparatedKey` with the prefix

product/log-service/handler/store.go
```diff
@@ -172,3 +177,74 @@ func HandleDelete(store store.Store) http.HandlerFunc {
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		st := time.Now()
+		h := w.Header()
+		h.Set("Access-Control-Allow-Origin", "*")
+		ctx := r.Context()
+
+		accountID := r.URL.Query().Get(accountIDParam)
+		prefix := r.URL.Query().Get(usePrefixParam)
+
+		filepathSuffix := "logs.zip"
```
vistaarjuneja, line 191: Move this outside this function and keep as a var
var (
filePathSuffix string = "logs.zip" )

vistaarjuneja (dismissed):

product/log-service/handler/store.go
rafaelmart1ns: It was changed.

vistaarjuneja (approved):

rafaelmart1ns:
trigger uts_JDK11

lasdox (approved):

PR #44233, opened 2023-02-22
fix: [DEL-6003]: Properly add user 1001

mdmilic:
## Describe your changes
USER docker command automatically creates user if it doesn't exist, but doesn't name it or add it to the user database (/etc/passwd). User missing in the /etc/passwd was causing the issue for SSH tasks. 

Fix is to first create user using a native ubi command (provided by the shadow-utils library) and then we can set it in Dockerfile (at which point docker doesn't create it)

Merged multiple consecutive RUN commands to reduce the image size. 

Testing Done:
* Fix tested by sales engs
* Verified again image works

## Checklist
- [x] I've documented the changes in the PR description.
- [x] I've tested this change either in PR or local environment.
- [ ] If hashcheck failed I followed [the checklist](https://harness.atlassian.net/wiki/spaces/DEL/pages/21016838831/PR+Codebasehash+Check+merge+checklist) and updated this PR description with my answers to make sure I'm not introducing any breaking changes.

## Comment Triggers
<details>
  <summary>Build triggers</summary>

- Feature build: `trigger feature-build`
- Immutable delegate `trigger publish-delegate`
</details>

<details>
  <summary>PR Check triggers</summary>

You can run multiple PR check triggers by comma separating them in a single comment. e.g. `trigger ti0, ti1`

- Compile: `trigger compile`
- CodeformatCheckstyle: `trigger checkstylecodeformat`
  - CodeFormat: `trigger codeformat`
  - Checkstyle: `trigger checkstyle`
- MessageMetadata: `trigger messagecheck`
- File-Permission-Check: `trigger checkpermission`
- Recency: `trigger recency`
- BuildNumberMetadata: `trigger buildnum`
- PMD: `trigger pmd`
- Copyright Check: `trigger copyrightcheck`
- Feature Name Check: `trigger featurenamecheck`
- UnitTests-ALL: `trigger utAll`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- FunctionalTest1: `trigger ft1`
- FunctionalTest2: `trigger ft2`
- CodeBaseHash: `trigger codebasehash`
- CodeFormatCheckstyle: `trigger checkstylecodeformat`
- SonarScan: `trigger ss`
- GitLeaks: `trigger gitleaks`
- Trigger all Checks: `trigger smartchecks`
</details>

## PR check failures and solutions
[...]

PR #50007, opened 2023-07-02
feat: [IE-175]: Single docker file for building Platform-service in harness-core

Vbhatia20:
## Describe your changes
https://app.harness.io/ng/account/vpCkHKsDSxK9_KYfjCTMKA/ci/orgs/default/projects/FEATUREBUILDS/pipelines/MultiStage_Docker_Build/executions/MOeI8H73Ro-qBO7wpDK3FA/pipeline?storeType=INLINE

Follow up ticket to be taken care : https://harness.atlassian.net/browse/IE-292

Image size earlier : 958MB
Image size now : 958MB
Build Time earlier :11mins
Build Time now : 13mins 

## Checklist

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/50007)


williamwissemann (approved):

Vbhatia20:
trigger uts

williamwissemann (approved):

bazelrc.remote
```diff
@@ -2,7 +2,7 @@ import bazelrc.common
 
 # Add properties that are specific for ci build system
 
-startup --output_base=/tmp
+#startup --output_base=/tmp
```
diptiman-adak, line 5: This flag was specific to Harness CI. Changing the output base to avoid space and inode exhaustion issue. However running the build within docker container might not need it.
But if we are using volume mount we need to use how this behaves. It might have some issues going forward.

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
```diff
@@ -0,0 +1,56 @@
...
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/config.yml /opt/harness/
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/classpath_metadata.json /opt/harness/
+
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/scripts /opt/harness
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/default.jfc /opt/harness/
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/profile.jfc /opt/harness/
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/profile.jfc /opt/harness/
+COPY --chown=65534:65534 --from=builder /app/dockerization/base-images/apm/*.sh /opt/harness
+
+RUN chmod +x /opt/harness/*.sh
```
diptiman-adak, line 40: This will bloat-up the docker image size. It will create duplicate layers. 
COPY --chown=65534:65534 use like this

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
```diff
@@ -0,0 +1,56 @@
...
+
+RUN mkdir -p /app
+ARG GCP_KEY
+COPY . /app
+
+WORKDIR /app
+
+ENV BUILD_PURPOSE=DEVELOPMENT
+
+RUN chmod +x \
```
diptiman-adak, line 16: This section should be avoided by mounting machine dir to docker.

platform-service/build/feature_build.sh
```diff
@@ -5,7 +5,7 @@
 # https://polyformproject.org/wp-content/uploads/2020/05/PolyForm-Free-Trial-1.0.0.txt.
 
 set -ex
-
+export GCP_KEY="$1"
```
diptiman-adak, line 8: @Vbhatia20 we should be see this GCP does not get exposed. Please inspect the image to confirm if we pass the key during the build it don't store it into layers. We had issues previously with this

diptiman-adak (changes requested):

diptiman-adak:
Two things I want to see which can be impacted due to this change.
1. Build time
2. Docker image size. 
Can you add the supporting data for these two areas

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
Vbhatia20: this is removed

platform-service/build/feature_build.sh
williamwissemann: As long as the script is only used in the `builder` target and isn't passed to the `base` target then the secret won't show up in the layers. It may be worth adding some comments about being careful with adding secrets to the base target

williamwissemann: @Vbhatia20 While the secret shouldn't end up in the `base` targets layers, we should ensure the build args aren't cached inside the container. There is a script on this confluence page that should detect them if they are:

https://harness.atlassian.net/wiki/spaces/SC/pages/21370306573/Scanning+Docker+Images+with+Gitleaks

You might want to explore to passing that secret in via the `--secret` flag ( docker build --secret )

Vbhatia20: I scanned and no git leaks found.

Vbhatia20:
trigger uts

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
```diff
@@ -0,0 +1,55 @@
+# Copyright 2022 Harness Inc. All rights reserved.
+# Use of this source code is governed by the PolyForm Free Trial 1.0.0 license
+# that can be found in the licenses directory at the root of this repository, also available at
+# https://polyformproject.org/wp-content/uploads/2020/05/PolyForm-Free-Trial-1.0.0.txt.
+
+FROM us.gcr.io/platform-205701/cie-agent-harness-core-jdk11:latest as builder
+
+RUN mkdir -p /app
+ARG GCP_KEY
```
mdmilic, line 9: Generally you can't have secrets as args, args are visible in the final image. Since this is not a final stage though, not sure if someone could extract the GCP_KEY or not.

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
```diff
@@ -0,0 +1,55 @@
...
+# Add the capsule JAR and config.yml
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/platform-service-capsule.jar /opt/harness/
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/keystore.jks /opt/harness/
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/config.yml /opt/harness/
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/classpath_metadata.json /opt/harness/
+
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/scripts /opt/harness
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/default.jfc /opt/harness/
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/profile.jfc /opt/harness/
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/profile.jfc /opt/harness/
```
mdmilic, line 37: duplicated

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
```diff
@@ -0,0 +1,55 @@
...
+RUN chmod +x /opt/harness/*.sh
+CMD [ "/opt/harness/run.sh" ]
+
+############################ON PREM#########################
+FROM base as onprem
+
+RUN /opt/harness/inject-onprem-apm-bins-into-dockerimage.sh && rm /opt/harness/inject-onprem-apm-bins-into-dockerimage.sh
+USER root
+RUN microdnf remove tar unzip gzip
+USER 65534
```
mdmilic, line 48: You should create the user properly. Take a look at this delegate PR and associated bug caused by this #44233

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
```diff
@@ -0,0 +1,55 @@
...
+COPY --chown=65534:65534 --from=builder /app/dockerization/base-images/apm/*.sh /opt/harness
+RUN chmod +x /opt/harness/*.sh
+CMD [ "/opt/harness/run.sh" ]
+
+############################ON PREM#########################
+FROM base as onprem
+
+RUN /opt/harness/inject-onprem-apm-bins-into-dockerimage.sh && rm /opt/harness/inject-onprem-apm-bins-into-dockerimage.sh
+USER root
+RUN microdnf remove tar unzip gzip
```
mdmilic, line 47: After running microdnf commands you should also cleanup the caches (to reduce the size)
```
  && microdnf clean all \
  && rm -rf /var/cache/yum \
```

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
```diff
@@ -0,0 +1,55 @@
...
+    /app/scripts/bazel/generate_credentials.sh \
+    && /app/scripts/bazel/generate_credentials.sh \
+    && /app/platform-service/build/feature_build.sh $GCP_KEY \
+    && /app/platform-service/build/build_jar.sh \
+    && /app/platform-service/build/build_dist.sh
+
+FROM us.gcr.io/platform-205701/ubi/ubi-java:8.8 as base
+
+# Add the capsule JAR and config.yml
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/platform-service-capsule.jar /opt/harness/
```
mdmilic, line 29: Checkout these redhat best practices around writing Dockerfiles with UBI as a base OS. [Depending if needed] one of them is to use root user group (e.g. --chown=65534:0). If not specified, when the container runs in openshift (e.g. in SMP), it will only have access to the files in root user group. This is because OpenShift for security reasons ignores the user:group you specified in your `USER` command and runs the container with a <random_uuid>:0 (i.e.  userId is random number in range 1001-65534 and user group is 0)

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
```diff
@@ -0,0 +1,55 @@
...
+############################ON PREM#########################
+FROM base as onprem
+
+RUN /opt/harness/inject-onprem-apm-bins-into-dockerimage.sh && rm /opt/harness/inject-onprem-apm-bins-into-dockerimage.sh
+USER root
+RUN microdnf remove tar unzip gzip
+USER 65534
+
+############################SAAS#########################
+FROM base as saas
```
mdmilic, line 51: I'm assuming you would want to build either SaaS or SMP in a single execution. Note that this could depend on the docker builder and could have negative effect on your build performance. Legacy builder would build all stages in the multi stage Dockerfile regardless if they are needed/requested. [BuildKit](https://docs.docker.com/build/buildkit/) would only build stages needed to build the stage you requested.

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
```diff
@@ -0,0 +1,55 @@
...
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/scripts /opt/harness
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/default.jfc /opt/harness/
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/profile.jfc /opt/harness/
+COPY --chown=65534:65534  --from=builder /app/dist/platform-service/profile.jfc /opt/harness/
+COPY --chown=65534:65534 --from=builder /app/dockerization/base-images/apm/*.sh /opt/harness
+RUN chmod +x /opt/harness/*.sh
+CMD [ "/opt/harness/run.sh" ]
+
+############################ON PREM#########################
+FROM base as onprem
```
mdmilic, line 43: Maybe rename to smp as per new naming convention.

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
```diff
@@ -0,0 +1,55 @@
...
+USER root
+RUN microdnf remove tar unzip gzip
+USER 65534
+
+############################SAAS#########################
+FROM base as saas
+
+RUN /opt/harness/inject-saas-apm-bins-into-dockerimage.sh && rm -rf /opt/harness/inject-saas-apm-bins-into-dockerimage.sh
+
+USER root
```
mdmilic, line 55: Do we want to run as root in SaaS

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
Vbhatia20: added these, but image size still same.

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
Vbhatia20: mounting will not work here

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
Vbhatia20: we need to build both and depending on target we will use it

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
diptiman-adak: Yes we need to run to debug issues. 
This has been requested by SRE team earlier.

Vbhatia20: we havent made any changes here

platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage
Vbhatia20: we haven't change anything here

Vbhatia20:
trigger uts

mdmilic (commented):
> Reviewed 3 of 5 files at r1, 1 of 2 files at r5, 1 of 3 files at r7, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/50007)* status: 1 change requests. 6. unresolved discussions. 0 of 1 approvals obtained. 

___
*[`platform-service/build/build_jar.sh` line 13 at r9](https://reviewable.io/reviews/harness/harness-core/50007#-NapvsXu45g3Q_41jgVQ:-NapvsXu45g3Q_41jgVR:beiihuy) ([raw file](https://github.com/harness/harness-core/blob/e10f11c75fecb07ce7a783a2fc046c69cc135cb3/platform-service/build/build_jar.sh#L13)):*
> ```Shell
> BAZEL_ARGUMENTS="--show_timestamps --announce_rc --experimental_convenience_symlinks=normal --remote_download_outputs=all --symlink_prefix=${BAZEL_DIRS}/"
> PLATFORM_SERVICE_MODULE="//platform-service/service:module_deploy.jar"
> bazel build --remote_cache=https://storage.googleapis.com/harness-bazel-cache --google_credentials=/tmp/storage_secret.json $PLATFORM_SERVICE_MODULE ${BAZEL_ARGUMENTS}
> ```

Any specific reason why you don't use bazelrc.remote here? This would implicitly use bazelrc.local instead (which is meant for localdev on mac - specifies mac specific flags, though without any negative effects if run on other platforms)
___
*[`platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage` line 16 at r2](https://reviewable.io/reviews/harness/harness-core/50007#gh-1253271446:-Napz-PA0SldpO60g_8_:bq7fljf) ([raw file](https://github.com/harness/harness-core/blob/2d3b7d859774c5f791d1e9472ee042a4062a40f7/platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage#L16)):*
<details><summary><i>Previously, Vbhatia20 (Vishal Bhatia) wrote…</i></summary><blockquote>

mounting will not work here
</blockquote></details>

The reason why you need to do chmod is because when files were originally created & committed to git, it was done with default permissions (I think 644). You can change permissions and commit that to git, then no need for everyone to do chmod. 
https://stackoverflow.com/questions/10516201/updating-and-committing-only-a-files-permissions-using-git-version-control
___
*[`platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage` line 48 at r5](https://reviewable.io/reviews/harness/harness-core/50007#gh-1255697325:-Naq09egDRCYBZyIoQhm:b1zozne) ([raw file](https://github.com/harness/harness-core/blob/ab92ce9788806a282eaf00b2b2f709b78382cf12/platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage#L48)):*
<details><summary><i>Previously, Vbhatia20 (Vishal Bhatia) wrote…</i></summary><blockquote>

we haven't change anything here
</blockquote></details>

I'm assuming you are copying this script from somewhere, it would be good to fix the issue anyway. 
___
*[`platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage` line 55 at r5](https://reviewable.io/reviews/harness/harness-core/50007#gh-1255722741:-Naq0k-ZEg6HGNomchcL:b-wn56bf) ([raw file](https://github.com/harness/harness-core/blob/ab92ce9788806a282eaf00b2b2f709b78382cf12/platform-service/build/container/Dockerfile-platform-service-cie-jdk-multistage#L55)):*
<details><summary><i>Previously, Vbhatia20 (Vishal Bhatia) wrote…</i></summary><blockquote>

we havent made any changes here
</blockquote></details>

This will be runAsUser for the session which actually runs the app (and by default other sessions, like exec/SSH). This is just a default and can be overridden when creating the pod/container. 
It would be better though to disable elevated privileges by default for the application session (if they are needed you should consciously configure them when creating pod and not get them by accident when you don't even configure anything). You can still use root when you exec into the pod if you really need root (which usually you shouldn't need anyway especially not in PROD).
NOTE: doesn't matter if you are copying these things from other scripts, stackoverflow or somewhere else, this is a new script, if you see something wrong you should fix it and not carry over the tech debt.

williamwissemann (dismissed):

mdmilic (approved):

williamwissemann (approved):

PR #51741, opened 2023-08-17
fix: [PL-40669]: check for oldest active token incase of default token not present

VikasMaddukuriHarness:
## Describe your changes
1. When delegate token is revoked, updated its name to include a suffix "_revoked".  This is reverted as this causes issue as we do not know the reason of delegate disconnection ,if the token was revoked or not .
2. Get the oldest active token incase of default token is not available when installing a delegate in ng
3. Modified the message which clearly mentions to create a token in case of no active token is available at a given scope when user tries to install a new delegate.
## Checklist

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/51741)


VikasMaddukuriHarness:
trigger uts

VikasMaddukuriHarness:
trigger ut1

VikasMaddukuriHarness:
trigger uts

400-rest/src/main/java/io/harness/service/impl/DelegateNgTokenServiceImpl.java
```diff
@@ -174,6 +178,28 @@ public DelegateTokenDetails getDelegateToken(String accountId, String name, bool
...
+    }
+    DelegateToken delegateToken = persistence.createQuery(DelegateToken.class)
+                                      .field(DelegateTokenKeys.accountId)
+                                      .equal(accountId)
+                                      .field(DelegateTokenKeys.owner)
+                                      .equal(owner)
+                                      .field(DelegateTokenKeys.status)
+                                      .equal(ACTIVE)
+                                      .order(DelegateTokenKeys.createdAt)
+                                      .get();
```
jennyjames21, line 195: Should this be list of delegateToken?

400-rest/src/main/java/io/harness/service/impl/DelegateNgTokenServiceImpl.java
```diff
@@ -124,6 +127,7 @@ public DelegateTokenDetails revokeDelegateToken(String accountId, String tokenNa
     UpdateOperations<DelegateToken> updateOperations =
         persistence.createUpdateOperations(DelegateToken.class)
             .set(DelegateTokenKeys.status, DelegateTokenStatus.REVOKED)
+            .set(DelegateTokenKeys.name, tokenName + REVOKED_TOKEN_SUFFIX)
```
jennyjames21, line 130: if you rename the token name, how will we identify if the delegate using revoked token or not, as delegate has the old token name and we always do token lookup based on the token name?
What's the advantage of renaming here?

jennyjames21 (changes requested):

400-rest/src/main/java/io/harness/service/impl/DelegateNgTokenServiceImpl.java
VikasMaddukuriHarness: We are just getting the oldest active token from the delegates collection. Why do we need a list here?

400-rest/src/main/java/io/harness/service/impl/DelegateNgTokenServiceImpl.java
VikasMaddukuriHarness: ok. Removing this suffix part and making the ttl of the revoked token to 3 days which is currently 30 days. But this will be applicable from now on only. Should we need a migration to delete the existing revoked delegate tokens?

920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java
```diff
@@ -45,7 +45,7 @@
 @FieldNameConstants(innerTypeName = "DelegateTokenKeys")
 @OwnedBy(HarnessTeam.DEL)
 public class DelegateToken implements PersistentEntity, UuidAware, CreatedAtAware, CreatedByAware, NameAndValueAccess {
-  public static final Duration TTL = ofDays(30);
+  public static final Duration TTL = ofDays(3);
```
xingchi-jin, line 48: we need to set valid until based on status. only REVOKED is 3 days

xingchi-jin (commented):
LG after a comment

920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java
VikasMaddukuriHarness: Yes this static value is used when revoking a token. We use this value to set the valid until field 
This is the place - 

link - https://github.com/harness/harness-core/blob/1789ce5f412d8926e4c4b33b7c4ed89d76d54b20/400-rest/src/main/java/io/harness/service/impl/DelegateNgTokenServiceImpl.java#L128

VikasMaddukuriHarness:
trigger uts

VikasMaddukuriHarness:
trigger uts

xingchi-jin (approved):

920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java
```diff
@@ -45,15 +45,17 @@
 @FieldNameConstants(innerTypeName = "DelegateTokenKeys")
 @OwnedBy(HarnessTeam.DEL)
 public class DelegateToken implements PersistentEntity, UuidAware, CreatedAtAware, CreatedByAware, NameAndValueAccess {
-  public static final Duration TTL = ofDays(30);
+  public static final Duration REVOKED_TOKEN_TTL = ofDays(3);
```
xingchi-jin, line 48: since the name nolonger unique index, do we still need this ?

920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java
```diff
@@ -45,15 +45,17 @@
...
 
   public static List<MongoIndex> mongoIndexes() {
     return ImmutableList.<MongoIndex>builder()
         .add(CompoundMongoIndex.builder()
                  .field(DelegateTokenKeys.accountId)
                  .field(DelegateTokenKeys.name)
```
xingchi-jin, line 54: shall we keep an index using tokenName, but not unique index

400-rest/src/main/java/io/harness/service/impl/DelegateNgTokenServiceImpl.java
```diff
@@ -271,26 +296,29 @@ public String getDefaultTokenName(DelegateEntityOwner owner) {
...
                                                 .project(DelegateTokenKeys.name, true)
                                                 .asList();
-    delegateTokenList.forEach(
-        delegateToken -> revokeDelegateToken(delegateToken.getAccountId(), delegateToken.getName()));
+    delegateTokenList.forEach(delegateToken
+        -> revokeDelegateToken(delegateToken.getAccountId(), delegateToken.getName(), delegateToken.getOwner()));
   }
 
-  private Query<DelegateToken> matchNameTokenQuery(String accountId, String tokenName) {
+  private Query<DelegateToken> matchNameTokenQuery(String accountId, String tokenName, DelegateEntityOwner owner) {
```
xingchi-jin, line 309: will this be a list of tokens in return ?

jennyjames21: This is a common function, it can be list when two token with the same name exists but in a different state. 
Some flow only requires Active token ones, but the authentication, we may need both

400-rest/src/main/java/io/harness/service/impl/DelegateNgTokenServiceImpl.java
```diff
@@ -271,26 +296,29 @@ public String getDefaultTokenName(DelegateEntityOwner owner) {
...
         .field(DelegateTokenKeys.name)
-        .equal(tokenName);
+        .equal(tokenName)
+        .field(DelegateTokenKeys.owner)
+        .equal(owner);
   }
 
   private DelegateTokenDetails getDelegateTokenDetails(DelegateToken delegateToken, boolean includeTokenValue) {
     DelegateTokenDetailsBuilder delegateTokenDetailsBuilder = DelegateTokenDetails.builder()
+                                                                  .uuid(delegateToken.getUuid())
```
jennyjames21, line 321: why this change?

920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java
jennyjames21: @xingchi-jin Mongo considers partial index so this case {accountId, name} order is already indexed.

@VikasMaddukuriHarness Once this goes to prod, please work with DBAs to remove the old index manually as we have seen recently it will not automatically get deleted

jennyjames21:
@VikasMaddukuriHarness When u test can you verify this use case
1) Create a token 'xxx' and start delegate with that token. Verify it's connected and HB coming throgh
2) Revoke the token "xxx", and verify delegate got disconnected
3) Create new token with the same name (xxx) with the same scope and start a new delegate with new token and verify it's connected
4) Do the same flow for different scope too

xingchi-jin:
also please test with:  register a delegate with a revoked token

mdmilic (changes requested):
> Reviewed 1 of 1 files at r2, 2 of 4 files at r4, 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/51741)* status: 12 file(s) to review. 2 change requests. 4. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 4462 at r5](https://reviewable.io/reviews/harness/harness-core/51741#-NeEDm3E7QqnvvsjpEbu:-NeEDm3E7QqnvvsjpEbv:bnt58cm) ([raw file](https://github.com/harness/harness-core/blob/c346279891b7d1da4e04ec297fb5d21748a20fbc/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L4462)):*
> ```Java
>     if (delegateTokenName.isPresent()) {
>       DelegateTokenDetails delegateTokenDetails =
>           delegateNgTokenService.getDelegateToken(accountId, null, delegateTokenName.get());
> ```

Why null here, that would mean account scope? The same on other places as well. We use this to register delegate to the appropriate scope that it's token belongs to. 
___
*[`920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java` line 54 at r5](https://reviewable.io/reviews/harness/harness-core/51741#gh-1322239794:-NeED-bu8ZMgSe7zzm4c:by2bewi) ([raw file](https://github.com/harness/harness-core/blob/c346279891b7d1da4e04ec297fb5d21748a20fbc/920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java#L54)):*

@xingchi-jin Mongo considers partial index so this case {accountId, name} order is already indexed.

@VikasMaddukuriHarness Once this goes to prod, please work with DBAs to remove the old index manually as we have seen recently it will not automatically get deleted
</blockquote></details>

I don't think we can change this index like this. I think our delegate auth mechanism assumes that the token is unique by name (so we cache JWT Token -> token.name), having multiple tokens with the same name across orgs/projs would cause collision and potentially authenticate delegate with to a wrong scope. Can you please double check those flows to make sure they would work with multiple tokens with the same name present in DB.

400-rest/src/main/java/io/harness/service/impl/DelegateNgTokenServiceImpl.java
VikasMaddukuriHarness: we are not returning the delegateToken UUID. This is required in case of delegate token crud using terraform. This does not impact any flow apart from the getCall for delegateToken

920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java
VikasMaddukuriHarness: you mean to keep the revoked token till 30 days. Ok I will revert that change

VikasMaddukuriHarness (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/51741)* status: Some of the checks are too old. 12 file(s) to review. 2 change requests. 7. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java` line 54 at r5](https://reviewable.io/reviews/harness/harness-core/51741#gh-1322239794:-NeMSr5q1aAqBWFzu_FQ:b-ssspyd) ([raw file](https://github.com/harness/harness-core/blob/c346279891b7d1da4e04ec297fb5d21748a20fbc/920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java#L54)):*

I don't think we can change this index like this. I think our delegate auth mechanism assumes that the token is unique by name (so we cache JWT Token -> token.name), having multiple tokens with the same name across orgs/projs would cause collision and potentially authenticate delegate with to a wrong scope. Can you please double check those flows to make sure they would work with multiple tokens with the same name present in DB.
</blockquote></details>

Yes @mdmilic we need to modify the auth mechanism as well in that case. We need to start hashing the token along with the scope details in it. If multiple scopes have same name, there will be a hash collision if we just use the token name to hash.
@jennyjames21 @xingchi-jin what do you think.

VikasMaddukuriHarness (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/51741)* status: Some of the checks are too old. 12 file(s) to review. 2 change requests. 7. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java` line 54 at r5](https://reviewable.io/reviews/harness/harness-core/51741#gh-1322239794:-NenIbkq0_D_pR8SHbKU:b52dnod) ([raw file](https://github.com/harness/harness-core/blob/c346279891b7d1da4e04ec297fb5d21748a20fbc/920-delegate-service-beans/src/main/java/io/harness/delegate/beans/DelegateToken.java#L54)):*
<details><summary><i>Previously, VikasMaddukuriHarness (VikasMaddukuriHarness) wrote…</i></summary><blockquote>

Yes @mdmilic we need to modify the auth mechanism as well in that case. We need to start hashing the token along with the scope details in it. If multiple scopes have same name, there will be a hash collision if we just use the token name to hash.
@jennyjames21 @xingchi-jin what do you think.
</blockquote></details>

@mdmilic I just went through the auth flow again. No where we are using the delegateTokenName to authenticate or store the hash.
The Jwt encrypted token coming from delegate does not depend on the tokenName. In the manager side, the hash which we store in cache also does not use the tokenName. So we are good in the token name changes in authentication of delegate.
I tested this throughly in pr env.

VikasMaddukuriHarness (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/51741)* status: Some of the checks are too old. 12 file(s) to review. 2 change requests. 9. unresolved discussions. 0 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java` line 4462 at r5](https://reviewable.io/reviews/harness/harness-core/51741#-NeEDm3E7QqnvvsjpEbu:-NenJ3c-EzghRbfxmsDv:b-g7yvtz) ([raw file](https://github.com/harness/harness-core/blob/c346279891b7d1da4e04ec297fb5d21748a20fbc/400-rest/src/main/java/software/wings/service/impl/DelegateServiceImpl.java#L4462)):*

Why null here, that would mean account scope? The same on other places as well. We use this to register delegate to the appropriate scope that it's token belongs to.
</blockquote></details>

I understand what you meant by this @mdmilic. Yes this seems to be a problem now.
Because now if we have a token name common for both org and account level, and now if we use that token to create to register a delegate, now delegate will randomly get into accountLevel scope only. Because we check for the delegate Token details using the delegate token name only. In this case even if we use the delegateToken of the org level, after doing the auth in manager side, we actually authenticate it with the org level token, after the auth when registering, we try to fetch the delegateToken details using the delegateToken name. From this token value, we try to fetch the delegate scope. 
The problem arises here. We fetch the details of the accountLevel token only because we are not sure which scope the delegate is trying to register to. 
We need to think of some other approach now to make this change.
@jennyjames21 @xingchi-jin FYI

PR #52680, opened 2023-09-11
fix: [PL-41151]: Add current delegate version to supported delegate version

jennyjames21:
## Describe your changes
Summary: Add account delegate version for both legacy and immutable for /supportedDelegateVersion API

Test
--
-Local Testing
-Unit test

## Checklist

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/52680)


jennyjames21:
trigger uts

jennyjames21:
trigger uts

jennyjames21:
trigger uts

jennyjames21:
trigger uts

jennyjames21:
trigger uts

mdmilic (changes requested):
> Reviewed 1 of 3 files at r2, 1 of 4 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/52680)* status: 5 file(s) to review. 1 change requests. 0 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java` line 177 at r4](https://reviewable.io/reviews/harness/harness-core/52680#-NeCwu9OCBzdm-TvJYYF:-NeCwu9OCBzdm-TvJYYG:bm273a9) ([raw file](https://github.com/harness/harness-core/blob/3c5d4e5233e4bb37f21c0d2797f6c07242277e85/400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java#L177)):*
> ```Java
>   }
> 
>   public SupportedDelegateVersion getSupportedDelegateVersion(String accountId) {
> ```

This would not work for SMP since they don't use rings. Better use one of the other available methods like `getDelegateImageTag` or `getImmutableDelegateImageTag`. Btw those two look to implement the same logic. You should squash them into one method. 
___
*[`400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java` line 188 at r4](https://reviewable.io/reviews/harness/harness-core/52680#-NeCumP25HcIOOr_LPit:-NeCumP25HcIOOr_LPiu:b-6iwju1) ([raw file](https://github.com/harness/harness-core/blob/3c5d4e5233e4bb37f21c0d2797f6c07242277e85/400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java#L188)):*
> ```Java
> 
>     return SupportedDelegateVersion.builder()
>         .latestSupportedVersion(immutableAccountOverrideVersion)
> ```

This should be latestAccountVersion I think, but better call it versionOverride or something along those lines. AccountVersion is confusing because the ring versions is accountVersion as well (i.e. based on accountId). 
Also, I'm not sure we should have a separate field for version override at all. This is customer facing API, if they set the override, they are not interested in the ring versions. IMO in case of override we should just set latestSupportedVersions to the override and minimalVersion to blank.

jennyjames21:
> > Reviewed 1 of 3 files at r2, 1 of 4 files at r3, 3 of 3 files at r4, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/harness/harness-core/52680)_ status: 5 file(s) to review. 1 change requests. 0 of 2 approvals obtained.
> 
> _[`400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java` line 177 at r4](https://reviewable.io/reviews/harness/harness-core/52680#-NeCwu9OCBzdm-TvJYYF:-NeCwu9OCBzdm-TvJYYG:bm273a9) ([raw file](https://github.com/harness/harness-core/blob/3c5d4e5233e4bb37f21c0d2797f6c07242277e85/400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java#L177)):_
> 
> > ```java
> >   }
> > 
> >   public SupportedDelegateVersion getSupportedDelegateVersion(String accountId) {
> > ```
> 
> This would not work for SMP since they don't use rings. Better use one of the other available methods like `getDelegateImageTag` or `getImmutableDelegateImageTag`. Btw those two look to implement the same logic. You should squash them into one method.
> 
> _[`400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java` line 188 at r4](https://reviewable.io/reviews/harness/harness-core/52680#-NeCumP25HcIOOr_LPit:-NeCumP25HcIOOr_LPiu:b-6iwju1) ([raw file](https://github.com/harness/harness-core/blob/3c5d4e5233e4bb37f21c0d2797f6c07242277e85/400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java#L188)):_
> 
> > ```java
> >     return SupportedDelegateVersion.builder()
> >         .latestSupportedVersion(immutableAccountOverrideVersion)
> > ```
> 
> This should be latestAccountVersion I think, but better call it versionOverride or something along those lines. AccountVersion is confusing because the ring versions is accountVersion as well (i.e. based on accountId). Also, I'm not sure we should have a separate field for version override at all. This is customer facing API, if they set the override, they are not interested in the ring versions. IMO in case of override we should just set latestSupportedVersions to the override and minimalVersion to blank.

Updated, moved getImmutableDelegateImageTag which handles oprem case, also removed account version and api kept at it's. I will keep the minimal one as customers already mite be using it, so riskier to remove now

jennyjames21:
trigger uts

gn-harness (approved):

jennyjames21:
trigger messagecheck

jennyjames21:
trigger gitleaks

jennyjames21:
trigger messagecheck

jennyjames21:
trigger messagecheck

mdmilic (changes requested):
> Reviewed 2 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/52680)* status: 5 file(s) to review. 1 change requests. 1 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java` line 184 at r5](https://reviewable.io/reviews/harness/harness-core/52680#-NeJn0zU5OdmgFyxjuEG:-NeJn0zU5OdmgFyxjuEH:b5alm5d) ([raw file](https://github.com/harness/harness-core/blob/374f6a7fc93eee64722140734904cc0a5f16ef96/400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java#L184)):*
> ```Java
>     return SupportedDelegateVersion.builder()
>         .latestSupportedVersion(latestSupportedDelegateImage)
>         .latestSupportedMinimalVersion(latestVersion.concat(".minimal"))
> ```

I'm not sure I understand the split/concat logic, but `latestSupportedDelegateImage` can be a version override including a custom image which doesn't necessarily have minimal counterpart. 
Also, version override doesn't even have to have `:` because if not provided, image tag defaults to latest (e.g. `harness/delegate`)

jennyjames21:
> > Reviewed 2 of 3 files at r5, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/harness/harness-core/52680)_ status: 5 file(s) to review. 1 change requests. 1 of 2 approvals obtained.
> 
> _[`400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java` line 184 at r5](https://reviewable.io/reviews/harness/harness-core/52680#-NeJn0zU5OdmgFyxjuEG:-NeJn0zU5OdmgFyxjuEH:b5alm5d) ([raw file](https://github.com/harness/harness-core/blob/374f6a7fc93eee64722140734904cc0a5f16ef96/400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java#L184)):_
> 
> > ```java
> >     return SupportedDelegateVersion.builder()
> >         .latestSupportedVersion(latestSupportedDelegateImage)
> >         .latestSupportedMinimalVersion(latestVersion.concat(".minimal"))
> > ```
> 
> I'm not sure I understand the split/concat logic, but `latestSupportedDelegateImage` can be a version override including a custom image which doesn't necessarily have minimal counterpart. Also, version override doesn't even have to have `:` because if not provided, image tag defaults to latest (e.g. `harness/delegate`)

I dont know what's the purpose/history of a minimal one either. Should we just remove then? Concern is it's API used by can be used by customers for configuring upgrades, so we can't be sure if it breaks any customer flow.

jennyjames21:
trigger uts

jennyjames21:
> > > Reviewed 2 of 3 files at r5, all commit messages.
> > > _[Reviewable](https://reviewable.io/reviews/harness/harness-core/52680)_ status: 5 file(s) to review. 1 change requests. 1 of 2 approvals obtained.
> > 
> > 
> > _[`400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java` line 184 at r5](https://reviewable.io/reviews/harness/harness-core/52680#-NeJn0zU5OdmgFyxjuEG:-NeJn0zU5OdmgFyxjuEH:b5alm5d) ([raw file](https://github.com/harness/harness-core/blob/374f6a7fc93eee64722140734904cc0a5f16ef96/400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java#L184)):_
> > > ```java
> > >     return SupportedDelegateVersion.builder()
> > >         .latestSupportedVersion(latestSupportedDelegateImage)
> > >         .latestSupportedMinimalVersion(latestVersion.concat(".minimal"))
> > > ```
> > 
> > 
> > I'm not sure I understand the split/concat logic, but `latestSupportedDelegateImage` can be a version override including a custom image which doesn't necessarily have minimal counterpart. Also, version override doesn't even have to have `:` because if not provided, image tag defaults to latest (e.g. `harness/delegate`)
> 
> I dont know what's the purpose/history of a minimal one either. Should we just remove then? The concern is it's API can be used by customers for configuring upgrades, so we can't be sure if it breaks any customer flow.

@mdmilic Added check If version override exists and don't have ':' then make the minimal version field empty

mdmilic (changes requested):
> Reviewed 1 of 3 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/52680)* status: 5 file(s) to review. 1 change requests. 0 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java` line 184 at r5](https://reviewable.io/reviews/harness/harness-core/52680#-NeJn0zU5OdmgFyxjuEG:-NeNUvzL2M8hgOLCW2DE:b-qsrka3) ([raw file](https://github.com/harness/harness-core/blob/374f6a7fc93eee64722140734904cc0a5f16ef96/400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java#L184)):*

@mdmilic Added check If version override exists and don't have ':' then make the minimal version field empty
</blockquote></details>

@jennyjames21, sorry I think this entire thing is all wrong (unrelated to what you are implementing). The business logic just doesn't make any sense and we are trying to stuff more things that don't make any sense in it. IMO we should just crate a new endpoint and deprecate this one (if someone is really using it). Also, your change still alters the existing behavior and has problems with version overrides.
The major problem I see with this endpoint is that it strips the repo information with the intention to just show the delegate version (i.e. `22.10.80500` not the delegate image version `harness/delegate:22.10.80500`). This also means that minimal version doesn't make any sense in this context (since we can't have minimal delegate version like `22.10.80500.minimal`, we can have minimal delegate image version like `harness/delegate:22.10.80500.minimal`), but also means that we can't show image overrides as we can't know the delegate version from the delegate image version like we do for our own images (e.g. `marko/delegate:custom-version` doesn't mean anything to us). 
So my proposal is create a new endpoint which has a more sane business logic and just returns the delegate image version that would have otherwise been used by the Upgrader to install it for your delegate (we can also throw in minimal delegate as well). So it would be like:
* If no override we populate the full delegate image tag + full minimal delegate image tag from ring (or from whatever is SMP using if minimal makes sense there). You could enhance the `getImmutableDelegateImageTag` to give back String + bool telling you if it was override or not. 
* If override just put that version in the latestSupported (no minimal override), make it null. 
* SMP considerations - TBD.

As for your current changes if you end up pursuing modifying existing endpoint (which I think is wrong and don't recommend) here are the issues:
* In case of ring version, your latest will be full image tag, your minimal will be just delegateVersion.minimal - latest used to be just delegateVersion (i.e. without repo information)
* In case of override you are still creating minimal version which doesn't make sense for override (overrides can still have `:`, but they don't have to) - minimal should here be the same as latest or empty
* In case of override you will return the full image tag (i.e. `marko/delegate:my-version`), but in case of ring you would return only delegate version (i.e. `22.10.80500`), so there is a difference in behavior <-- this is a major issue
* You should add following unit tests with real data (I was wrong to use custom data there, but it didn't matter before this change)
  * Ring image exists (e.g. `harness/delegate:22.10.80500), no override 
  * Ring image exists and override exists (e.g. 'marko/delegate:my-version)'
  * Ring image exists and override exists but without image tag (e.g. `marko/delegate`)

jennyjames21:
> > Reviewed 1 of 3 files at r5, 2 of 2 files at r6, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/harness/harness-core/52680)_ status: 5 file(s) to review. 1 change requests. 0 of 2 approvals obtained.
> 
> _[`400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java` line 184 at r5](https://reviewable.io/reviews/harness/harness-core/52680#-NeJn0zU5OdmgFyxjuEG:-NeNUvzL2M8hgOLCW2DE:b-qsrka3) ([raw file](https://github.com/harness/harness-core/blob/374f6a7fc93eee64722140734904cc0a5f16ef96/400-rest/src/main/java/io/harness/delegate/service/DelegateVersionService.java#L184)):_
> 
> Previously, jennyjames21 (Jenny James) wrote…
> @jennyjames21, sorry I think this entire thing is all wrong (unrelated to what you are implementing). The business logic just doesn't make any sense and we are trying to stuff more things that don't make any sense in it. IMO we should just crate a new endpoint and deprecate this one (if someone is really using it). Also, your change still alters the existing behavior and has problems with version overrides. The major problem I see with this endpoint is that it strips the repo information with the intention to just show the delegate version (i.e. `22.10.80500` not the delegate image version `harness/delegate:22.10.80500`). This also means that minimal version doesn't make any sense in this context (since we can't have minimal delegate version like `22.10.80500.minimal`, we can have minimal delegate image version like `harness/delegate:22.10.80500.minimal`), but also means that we can't show image overrides as we can't know the delegate version from the delegate image version like we do for our own images (e.g. `marko/delegate:custom-version` doesn't mean anything to us). So my proposal is create a new endpoint which has a more sane business logic and just returns the delegate image version that would have otherwise been used by the Upgrader to install it for your delegate (we can also throw in minimal delegate as well). So it would be like:
> 
> * If no override we populate the full delegate image tag + full minimal delegate image tag from ring (or from whatever is SMP using if minimal makes sense there). You could enhance the `getImmutableDelegateImageTag` to give back String + bool telling you if it was override or not.
> * If override just put that version in the latestSupported (no minimal override), make it null.
> * SMP considerations - TBD.
> 
> As for your current changes if you end up pursuing modifying existing endpoint (which I think is wrong and don't recommend) here are the issues:
> 
> * In case of ring version, your latest will be full image tag, your minimal will be just delegateVersion.minimal - latest used to be just delegateVersion (i.e. without repo information)
> * In case of override you are still creating minimal version which doesn't make sense for override (overrides can still have `:`, but they don't have to) - minimal should here be the same as latest or empty
> * In case of override you will return the full image tag (i.e. `marko/delegate:my-version`), but in case of ring you would return only delegate version (i.e. `22.10.80500`), so there is a difference in behavior <-- this is a major issue
> * You should add following unit tests with real data (I was wrong to use custom data there, but it didn't matter before this change)
>   
>   * Ring image exists (e.g. `harness/delegate:22.10.80500), no override
>   * Ring image exists and override exists (e.g. 'marko/delegate:my-version)'
>   * Ring image exists and override exists but without image tag (e.g. `marko/delegate`)

@mdmilic, 
1) Added new end point.
2) enhance getImmutableDelegateImageTag by adding minimal version, without touching existing usuage
3) i will add unit test once finalized the approach

120-ng-manager/src/main/java/io/harness/ng/core/delegate/client/DelegateNgManagerCgManagerClient.java
```diff
@@ -182,4 +182,8 @@ Call<RestResponse<String>> overrideDelegateImage(
   @GET("version/supportedDelegate")
   Call<RestResponse<SupportedDelegateVersion>> getPublishedDelegateVersion(
       @Query(NGCommonEntityConstants.ACCOUNT_KEY) @NotNull String accountIdentifier);
+
+  @GET("version/delegate-latest")
```
raghuAtWings, line 186: why new API?

we should just make a change so that when the api is called, we return the same image as we will return for upgrader which is account based.

raghuAtWings (approved):

120-ng-manager/src/main/java/io/harness/ng/core/delegate/client/DelegateNgManagerCgManagerClient.java
jennyjames21: the existing one is created specifically for ring-based versions so changing behavior may be risky as it's customer exposed api
For safer side, we creating new end point and deprecate other one eventually

jennyjames21: Also current one will not work for on-prem

raghuAtWings: @jennyjames21 no, the customer has no idea about ring, all they care about is what version is deployed for them. We should not add new API. 

Why will current one not work for on prem? if it doesn't then we need to fix that as well. In smp there are no rings and we should return what delegate version the smp release was signed off with,

PR #54737, opened 2023-10-27
fix: [PL-42248]: Support of custom image in version override

jennyjames21:
## Describe your changes
Summary:
- Ability to distinct custom images with delegateImageTag  in version override file
- For /supported-version api always returns latest harness override version if any else,  show ring version
- For upgrade, download ui, delegate listing page show image with highest priority on custom image  
    custom image(if available). > harness version override image(if available)  > Ring image

Test
-Unit test
- Devspace testing -TBD

## Checklist

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

### Rebase Instructions:
To rebase your branch onto the latest changes in the target branch, simply use the following command in a comment: `/rebase`

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/54737)


jennyjames21:
trigger uts

jennyjames21:
trigger uts

jennyjames21:
trigger uts

420-delegate-service/src/main/java/io/harness/delegate/beans/VersionOverrideType.java
```diff
@@ -11,4 +11,11 @@
...
 import io.harness.annotations.dev.OwnedBy;
 
-@OwnedBy(DEL) public enum VersionOverrideType { DELEGATE_JAR, WATCHER_JAR, DELEGATE_IMAGE_TAG, UPGRADER_IMAGE_TAG }
+@OwnedBy(DEL)
+public enum VersionOverrideType {
+  DELEGATE_JAR,
+  WATCHER_JAR,
+  DELEGATE_IMAGE_TAG,
+  UPGRADER_IMAGE_TAG,
+  DELEGATE_CUSTOM_IMAGE_TAG
```
mdmilic, line 20: Not sure I understand the why are we introducing this as a new type of override? Any kind of image override should take precedence over Ring version, but also this allows you to have two overrides for account breaking uniqueness constraint. What would be the distinction of this type of override VS the old one DELEGATE_IMAGE_TAG

mdmilic (changes requested):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/54737)* status: 7 file(s) to review. 1 change requests. 0 of 2 approvals obtained. 

___
*[`420-delegate-service/src/main/java/io/harness/delegate/beans/VersionOverrideType.java` line 20 at r3](https://reviewable.io/reviews/harness/harness-core/54737#gh-1376064469:-Ni0JcnR54-7DdqfbzwQ:b9y46mf) ([raw file](https://github.com/harness/harness-core/blob/742c0dc7d1cc47eacfef1b660aa4e9708fc6fd67/420-delegate-service/src/main/java/io/harness/delegate/beans/VersionOverrideType.java#L20)):*

@mdmilic we have a new requirement for a customer who wants to auto-upgrade their custom image.

So in their pipeline, they will fetch /latest-supported-version API to get the latest harness supported version for their account(Hot fixed version or Ring version). In this PR, I changed to get the latest supported version which will always return DELEGATE_IMAGE_TAG from version override, not a custom image if any.

Then they use /override-image-tag API to update their custom image. Here we update DELEGATE_CUSTOM_IMAGE_TAG. So there will be two entries if the hotfix version(harness override) and the custom version are available.

And getImmutableDelegateImageTag() will return Custom Image if available if not harness version override value and ring version if no Overrides

Currently, i don't see any uniqueness enforced for the account in the code. And there is no way to distinguish harness override versus custom image override

There are only 2 entries in prod2 and one in prod1 with custom images so adding migration to handle that. 
</blockquote></details>

Hey @jennyjames21, thanks for the clarification. I went through the ticket as well to better understand the requirements. 
My understand is that this is trying to solve a use case where customer uses version override, but needs a hotfix, so they don't want to deploy it to the entire fleet of delegates but to a smaller number or single delegate (e.g. so they try out the hotfix)? Is my understanding correct?
If I understood the root problem correctly, my personal opinion is that we shouldn't introduce this kind of change for it and should ask customer to disable/pause auto upgrades and try the hotfix, when they are happy with the hotfix we should deploy it to all their delegates. This is way simpler even for the customer than trying to create some weird automations (like reading one API do some checks then write to other API etc...), but is also an extremely rare use case. 

Even if we still want to make some improvement here, I don't think this is a good change to make. It would introduce a loot of complexity which is not really easy to understand and can lead to more issues for general use cases. 
Instead I would suggest you to actually implement **version overrides per org/proj/delegate** (and remove extra override type), which should solve this but also other use cases. I don't know the current state of the code, but when I was designing this feature initially I was anticipating the possibility for customer to have more granular configuration of their delegate versions, so the code should mostly be setup for it, you would just need to store extra data in the DB and modify read/write operations. Having greater granularity would keep our APIs and logic simple, but should give customers what they want (and should be easier for them to use). If you do go down this route you should probably pick one that this specific customer needs and don't need to implement all granularities (e.g. org/proj vs delegate-name).
As for uniqueness, I was sure I had a mongo lvl unique(accountId, overrideType), but doesn't seem to be. However the uniqueness is [ensured](https://github.com/harness/harness-core/blob/develop/400-rest/src/main/java/io/harness/datahandler/services/AdminDelegateVersionService.java#L54) in code by using upsert operation.

jennyjames21:
> > _[Reviewable](https://reviewable.io/reviews/harness/harness-core/54737)_ status: 7 file(s) to review. 1 change requests. 0 of 2 approvals obtained.
> 
> _[`420-delegate-service/src/main/java/io/harness/delegate/beans/VersionOverrideType.java` line 20 at r3](https://reviewable.io/reviews/harness/harness-core/54737#gh-1376064469:-Ni0JcnR54-7DdqfbzwQ:b9y46mf) ([raw file](https://github.com/harness/harness-core/blob/742c0dc7d1cc47eacfef1b660aa4e9708fc6fd67/420-delegate-service/src/main/java/io/harness/delegate/beans/VersionOverrideType.java#L20)):_
> 
> Previously, jennyjames21 (Jenny James) wrote…
> Hey @jennyjames21, thanks for the clarification. I went through the ticket as well to better understand the requirements. My understand is that this is trying to solve a use case where customer uses version override, but needs a hotfix, so they don't want to deploy it to the entire fleet of delegates but to a smaller number or single delegate (e.g. so they try out the hotfix)? Is my understanding correct? If I understood the root problem correctly, my personal opinion is that we shouldn't introduce this kind of change for it and should ask customer to disable/pause auto upgrades and try the hotfix, when they are happy with the hotfix we should deploy it to all their delegates. This is way simpler even for the customer than trying to create some weird automations (like reading one API do some checks then write to other API etc...), but is also an extremely rare use case.
> 
> Even if we still want to make some improvement here, I don't think this is a good change to make. It would introduce a loot of complexity which is not really easy to understand and can lead to more issues for general use cases. Instead I would suggest you to actually implement **version overrides per org/proj/delegate** (and remove extra override type), which should solve this but also other use cases. I don't know the current state of the code, but when I was designing this feature initially I was anticipating the possibility for customer to have more granular configuration of their delegate versions, so the code should mostly be setup for it, you would just need to store extra data in the DB and modify read/write operations. Having greater granularity would keep our APIs and logic simple, but should give customers what they want (and should be easier for them to use). If you do go down this route you should probably pick one that this specific customer needs and don't need to implement all granularities (e.g. org/proj vs delegate-name). As for uniqueness, I was sure I had a mongo lvl unique(accountId, overrideType), but doesn't seem to be. However the uniqueness is [ensured](https://github.com/harness/harness-core/blob/develop/400-rest/src/main/java/io/harness/datahandler/services/AdminDelegateVersionService.java#L54) in code by using upsert operation.

What i meant by Hotfix is harness doing hotfix(have an entry in version override with hotfix version) for that account includes all delegates in the account.

raghuAtWings (dismissed):

jennyjames21:
> > Reviewed 3 of 4 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/harness/harness-core/54737)_ status: 7 file(s) to review. 1 change requests. 1 of 2 approvals obtained.
> 
> _[`390-db-migration/src/main/java/io/harness/migrations/all/UpdateOverrideTypeToCustomDelegateImage.java` line 42 at r4](https://reviewable.io/reviews/harness/harness-core/54737#-Ni9siXl6w0ABQwdSVtx:-Ni9siXl6w0ABQwdSVty:b1w4n7e) ([raw file](https://github.com/harness/harness-core/blob/dff9252a1a06a9de64f1e2ead2b3ca2ec4dc3df0/390-db-migration/src/main/java/io/harness/migrations/all/UpdateOverrideTypeToCustomDelegateImage.java#L42)):_
> 
> > ```java
> >       while (iterator.hasNext()) {
> >         VersionOverride versionOverride = iterator.next();
> >         if (isNotEmpty(versionOverride.getVersion()) && !versionOverride.getVersion().startsWith("harness/delegate")) {
> > ```
> 
> You should do a check if the override type is DELEGATE_IMAGE_TAG (e.g. legacy delegate.jar and watcher.jar overrides would get updated here)
> 
> _[`400-rest/src/main/java/io/harness/delegate/resources/DelegateVersionOverrideInternalResource.java` line 66 at r4](https://reviewable.io/reviews/harness/harness-core/54737#-Ni9tQPQA2RYN2uY1GZl:-Ni9tQPQA2RYN2uY1GZm:b-w9f1ye) ([raw file](https://github.com/harness/harness-core/blob/dff9252a1a06a9de64f1e2ead2b3ca2ec4dc3df0/400-rest/src/main/java/io/harness/delegate/resources/DelegateVersionOverrideInternalResource.java#L66)):_
> 
> > ```java
> >       @QueryParam("validTillNextRelease") @NotNull final Boolean validTillNextRelease,
> >       @QueryParam("validForDays") @NotNull final int validForDays) {
> >     return new RestResponse<>(adminDelegateVersionService.setCustomDelegateImageTag(
> > ```
> 
> The same API is called even when we want to override a version for customer (with harness version). Also customer might not necessarily set the override to a custom image, they could just use a public delegate image (in which case we should not consider it as custom IMO - e.g. they could set it to something which fixes an issue they have). Finally, not sure if we want to handle harness images from a private repo (i.e. they wouldn't start with `harness/delegate`, but are harness images, not custom). I think they should be considered as custom images for the purposes of our `/latest-supported-version` API, but wanted to bring it up.
> 
> _[`420-delegate-service/src/main/java/io/harness/delegate/beans/VersionOverrideType.java` line 20 at r3](https://reviewable.io/reviews/harness/harness-core/54737#gh-1376064469:-Ni9tkHV8nxJY8xDkM4d:b-ulmsaq) ([raw file](https://github.com/harness/harness-core/blob/742c0dc7d1cc47eacfef1b660aa4e9708fc6fd67/420-delegate-service/src/main/java/io/harness/delegate/beans/VersionOverrideType.java#L20)):_
> 
> Previously, mdmilic (Marko Milic) wrote…
> After more offline discussion, my understand was not correct. We want the `/latest-supported-version` API endpoint to be used only so customers can discover which delegate tag versions are available to them (e.g. so they can use that as a base for their custom image). It doesn't make sense for this endpoint to return custom image tag based on that description. The upgrader version-check endpoint on the other hand always needs to return the latest thing set (custom override or otherwise).

check is already  as part of query above, line before

jennyjames21:
*[`400-rest/src/main/java/io/harness/delegate/resources/DelegateVersionOverrideInternalResource.java` line 66 at r4](https://reviewable.io/reviews/harness/harness-core/54737#-Ni9tQPQA2RYN2uY1GZl:-NiKkk3RCFw6QFp9R9u3:bpumzsi) ([raw file](https://github.com/harness/harness-core/blob/dff9252a1a06a9de64f1e2ead2b3ca2ec4dc3df0/400-rest/src/main/java/io/harness/delegate/resources/DelegateVersionOverrideInternalResource.java#L66)):*

check is already  as part of query above, line before 
</blockquote></details>

Thinking more about this problem, I think you should not change anything related to setting the version override (i.e. new override type or anything on the set path). That would get very complex very easy. Instead I think you should just filter out the versions you are not interested in when invoking `/latest-supported-version` (e.g. if it doesn't start with `harness/delegate` - don't return it or something simple like that). This change would be way simpler I think.

What if the custom image name starts with harness/delegate?

We need to apply hierarchy on 2 places  for showing version-override 
1) latest-supported version :  Output harness version override
2) Other places like download UI, show version etc: Output custom image version override

mdmilic (commented):
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/54737)* status: Some of the checks are too old. 7 file(s) to review. 1 change requests. 1 of 2 approvals obtained. 

___
*[`400-rest/src/main/java/io/harness/delegate/resources/DelegateVersionOverrideInternalResource.java` line 66 at r4](https://reviewable.io/reviews/harness/harness-core/54737#-Ni9tQPQA2RYN2uY1GZl:-NiLu1QIBaXD2hdWiyp6:b-kuqt07) ([raw file](https://github.com/harness/harness-core/blob/dff9252a1a06a9de64f1e2ead2b3ca2ec4dc3df0/400-rest/src/main/java/io/harness/delegate/resources/DelegateVersionOverrideInternalResource.java#L66)):*

Thinking more about this problem, I think you should not change anything related to setting the version override (i.e. new override type or anything on the set path). That would get very complex very easy. Instead I think you should just filter out the versions you are not interested in when invoking `/latest-supported-version` (e.g. if it doesn't start with `harness/delegate` - don't return it or something simple like that). This change would be way simpler I think.

What if the custom image name starts with harness/delegate?

We need to apply hierarchy on 2 places  for showing version-override 
1) latest-supported version :  Output harness version override
2) Other places like download UI, show version etc: Output custom image version override
</blockquote></details>

After more discussion. The case which would not work if we use just one override type:
* We set a hotfix version for customer (`DELEGATE_IMAGE_TAG`)
* Customer calls /latest-supported-version and builds custom image
* Customer sets the override (so that their upgrader picks that up
* /latest-supported-version would now return ring image.

To make the two override type solution work we need to do a check if someone is setting our own public image or something else and set the override type based on that.

jennyjames21:
trigger uts

mdmilic (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/harness/harness-core/54737)* status: 7 file(s) to review. 1 of 2 approvals obtained.

jennyjames21:
trigger uts

jennyjames21:
trigger messagecheck

PR #55575, opened 2023-11-17
feat: [PL-41193]: Infra Cleanup request

mdmilic:
## Describe your changes
Expose Infra Cleanup API for bijou flows and wire it through to the K8S Runner cleanup stage. Send cleanup response back.

Tested by triggering the APIs with Postman

## Checklist

## Comment Triggers
<details>

- 
  Specific builds
  - `trigger manager`
  - `trigger dms`
  - `trigger ng_manager`
  - `trigger cvng `
  - `trigger cmdlib`
  - `trigger template_svc`
  - `trigger event_server`
</details>

<details>

- Compile: `trigger compile`
- Recency: `trigger recency`
- PMD: `trigger pmd`
- UnitTests-0: `trigger ut0`
- UnitTests-1: `trigger ut1`
- UnitTests-2: `trigger ut2`
- UnitTests-3: `trigger ut3`
- UnitTests-4: `trigger ut4`
- UnitTests-5: `trigger ut5`
- UnitTests-6: `trigger ut6`
- UnitTests-7: `trigger ut7`
- UnitTests-8: `trigger ut8`
- UnitTests-9: `trigger ut9`
- SonarScan: `trigger ss`
- Go Build: `trigger gobuild`
</details>

### Rebase Instructions:
To rebase your branch onto the latest changes in the target branch, simply use the following command in a comment: `/rebase`

- - -
This change is [Reviewable](https://reviewable.io/reviews/harness/harness-core/55575)


mdmilic:
trigger uts

mdmilic:
trigger uts

mdmilic:
trigger ut5, ut6, ut7, ut8, ut9

mdmilic:
trigger uts

mdmilic:
trigger uts

mdmilic:
trigger uts

mdmilic:
trigger ut5

260-delegate/src/main/java/io/harness/delegate/service/handlermapping/handlers/CleanupHandler.java
```diff
@@ -7,15 +7,49 @@
...
+      response = CleanupInfraResponse.newBuilder().setResponseCode(ResponseCode.RESPONSE_OK).build();
+    } catch (Exception e) {
+      log.error("Cleanup infra by runner {} failed with exception ", runner, e);
+      response = CleanupInfraResponse.newBuilder()
+                     .setResponseCode(ResponseCode.RESPONSE_FAILED)
+                     .setErrorMessage("Failed to cleanup the infra")
+                     .build();
+    }
+    final var call = managerClient.sendCleanupInfraResponse(
+        taskPayload.getId(), context.get(Context.DELEGATE_ID), context.get(Context.ACCOUNT_ID), response);
```
xingchi-jin, line 51: should here use taskPayload.getExecutionInfraId() ?

420-delegate-service/src/main/java/io/harness/grpc/scheduler/ScheduleTaskServiceGrpcImpl.java
```diff
@@ -104,45 +113,84 @@ public void initTask(final SetupExecutionInfrastructureRequest request,
...
+      log.error("accountId is empty");
+      responseObserver.onError(Status.INVALID_ARGUMENT.withDescription("accountId is mandatory").asRuntimeException());
+    }
+
+    if (StringUtils.isEmpty(request.getInfraRefId())) {
+      log.error("Infra ref id is empty for account {}", request.getAccountId());
       responseObserver.onError(
           Status.INVALID_ARGUMENT.withDescription("infra_ref_id is mandatory").asRuntimeException());
       return;
     }
```
xingchi-jin, line 155: before sending request to delete the actual LE, i think we should mark the executionInfraLocation as "tainted". so any possible execute will not be able to use the infra immediately.

420-delegate-service/src/main/java/io/harness/executionInfra/ExecutionInfraLocation.java
```diff
@@ -36,6 +36,7 @@
 @FieldNameConstants(innerTypeName = "ExecutionInfraLocationKeys")
 public class ExecutionInfraLocation implements PersistentEntity, UuidAware, CreatedAtAware, UpdatedAtAware {
   @Setter @Id private String uuid;
+  private final String accountId;
```
xingchi-jin, line 39: add a field as tainted, it is set true when receiving the clean up request

420-delegate-service/src/main/java/io/harness/grpc/scheduler/ScheduleTaskServiceGrpcImpl.java
mdmilic: No need to worry about that (at least not for now), this is not our responsibility (right now). Pipeline service is orchestrating the init/execute/cleanup events and they are dealing with this problem that they issue cleanup after all the other ones are handled. If it proves to become a problem we can address it later on.

260-delegate/src/main/java/io/harness/delegate/service/handlermapping/handlers/CleanupHandler.java
mdmilic: No, this needs to be ID of the cleanup task itself, we use it to find the task in the DB. But I think I need to pass infraRefId in addition to that, because I'm misusing this ID on the backend side on one place. I could get infraRefId from the Task object, but I think it's more correct to pass it as a query param. Thanks for pointing it out, will add another param.

920-delegate-service-beans/src/main/proto/io/harness/delegate/schedule_task.proto
```diff
@@ -48,12 +59,15 @@ message SchedulingConfig {
...
   string runner_type = 3;
 
-  string account_id = 4;
-  string org_id = 5;
-  string project_id = 6;
+  // TODO: Remove orgId and projectId from SchedulingConfig.
+  //  These are only used for fetching the secrets when we have scopedSecretId and don't really relate to scheduling.
+  //  Also, if we already have scopedId, we should not need these ids, we should update our APIs if needed instead.
+  // NOTE: Don't use these fields for any other purpose.
+  string org_id = 4;
```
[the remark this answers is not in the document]
mdmilic: One of my principal eng had a saying `good intentions don't work (mechanisms do)` 😃, I think it's a famous quote from Jeff Bezos. 
The problem is that *`as long as`* is not going to work. If you put it in the API, you can't control the usage, someone will start using it for different purposes in couple of months or even completely misusing it and it's going to be hard to go back (this is why I insist on lean API, API is our contract which is super hard to change ones it's live/signed and there is less possibilities for misuse if it's smaller). Also, you say you don't want it to be used for **`Scheduling`** decisions, but you put it inside **`SchedulingConfig`** which is used for those scheduling decisions 😄 . 
AccountId is always needed (even if it's not really technically needed) because we use it for customer isolation. That's why Harness global policy is that we always have accountId in the API path or Query Param for instance for every single API. Also, you can't make DB queries if you don't supply accountId, so I've just taken accountId to top level, because I don't always need SchedulingConfig, but need accountId (also, it's not really related to scheduling). 

I don't have much against using orgId/projId when they are needed (and they are sometimes and we as Harness do plan to continue using them, just differently), but the main problem here is that the API is disconnected. To fetch the secret you need scopedId + orgId + projId. Your scopedId is in Secret object deeply nested in your request, but your orgId & projId are in the top level SchedulingConfig (which is also available even when you don't have Secrets and even for other types of requests which are not related to dealing with secrets). So, to address this you could move orgId/projId to a secret object next to scopedId (this is your mechanism to prevent abuse from the quote above), However IMO the real issue is that scopedId itself is broken. 
If API is already asking me for a scopedId (meaning some service that understands scopes have created it) then that ID should be enough to figure out the scopes and all. If I need to understand the scopes and provide the orgId/projId, then what's the point of the scopedId, why not just providing standard secretId and the receiver side figures out the scopes. This is telling me that the actual fix is to entirely get rid of orgId & projId and fix the SM APIs and scopedId, so you only need one ID and nothing else. When we do find real usages for orgId/projId we can introduce them in a proper manner (i.e. put them in the right place in the request). 
As for SetupAbstractions, that's just a bad concept and we have a chance to not repeat the past mistakes. SetupAbstractions is just a bucket we use to put all the random stuff you don't know where to put.

mdmilic:
trigger uts

mdmilic:
trigger ut4

920-delegate-service-beans/src/main/proto/io/harness/delegate/schedule_task.proto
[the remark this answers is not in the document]
xingchi-jin: i agree org/proj id is not needed for scheduleConfig
i also agree with the simpler approach, where one id identifies everything. Unfortunately it's not the case in other Harness components, thus scopeId has two benefits
1. it's compatible with the current way pipeline works
2. it's simpler. because the secret named by the scopedId is mounted to the task container. It's easier for the tasks to know how to read the secret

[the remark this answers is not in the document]
xingchi-jin: it's not blocking comment, i'm ok with the change

xingchi-jin (approved):

PR #55637, opened 2023-11-20
fix: [SMP-2266]: PR to remove hardcoded values

arya-harness:


charts/delegate-proxy/values.yaml
```diff
@@ -74,54 +68,49 @@ securityContext:
...
 virtualService:
   annotations: {}
 ingress:
-  annotations: {}
\ No newline at end of file
+  annotations: {}
+configData:
+  proxy:
+    conf: |-
+      server { root /www/data;proxy_http_version 1.1;
```
arya-harness, line 115: magic of yq !!!

swapnil-harness (approved):

